PR #4, opened 2022-09-20
Add Clean Queue

gdsgj:
This adds an Amazon Simple Queuing Service queue that can be used for filtered events to be passed to the Clean Lambda


gareth-smyth-gov (commented):
I think we need to be more specific with our actions and resources instead of using the '*' wildcard.

PR #53, opened 2022-10-11
BTM-62: Updating SQS to use AWS SDK V3.

gareth-smyth-gov:


src/handlers/clean/handler.ts
```diff
@@ -39,7 +37,7 @@ export const handler = async (event: SQSEvent) => {
 
 async function cleanRecord(record: SQSRecord) {
   const bodyObject = JSON.parse(record.body);
-  console.log("body " + bodyObject);
+  console.log("body " + record.body);
```
markschnitziusgovuk, line 40: Pretty sure this will just log "[Object object]".

markschnitziusgovuk (approved):

PR #55, opened 2022-10-12
Btm 107 tests ci

NKannan14:


integration_tests/setup/testConstants.ts
```diff
@@ -1,4 +1,4 @@
-const dynamoDbTable = "di-btm-StorageTable-WHKGC5YIMCZF";
+const dynamoDbTable = "di-btm-StorageTable-D6OI93N75URV";
```
gareth-smyth-gov, line 1: You could use the `ListTablesCommand` from the dynamodb client to get all the tables then filter so this doesn't need to be updated each run.

integration_tests/tests.ts
```diff
@@ -1,19 +1,25 @@
...
 
-//below E2E tests is marked as skipped as still development in progress
-describe.skip("E2E tests", () => {
+describe("E2E tests", () => {
   beforeAll(async () => {
     snsResponse = await publishSNS(snsParams);
     expect(snsResponse).toHaveProperty("MessageId");
   });
   test("Publish sns message and expect message to reach dynamoDB ", async () => {
+    await delay(5000);
```
gareth-smyth-gov, line 18: It probably doesn't make a lot of difference with the small number of tests we have but a polling system with a timeout might be better than delaying for a number of seconds - i.e. each time we want to check if something exists in the DB we check (say) every five seconds until it is true and time out after (say) thirty seconds.

integration_tests/setup/testConstants.ts
NKannan14: added ListTables function to get the table name

integration_tests/tests.ts
NKannan14: fixed, added while condition to check for the event in the DB with timeout

integration_tests/setup/testConstants.ts
gareth-smyth-gov: 👍

gareth-smyth-gov (approved):

PR #57, opened 2022-10-13
Btm 105 - Template yaml cleanup

gareth-smyth-gov:
* Splitting template yaml into different files for readability.
* Adding build step to recombine these files before sam commands.


NKannan14 (approved):
looks good to me

PR #84, opened 2022-10-26
Add VAT

gdsgj:
This adds automatic upload of a runtime config file for historical UK Value-Added Tax rates to Amazon Simple Storage Service. This is done with an Amazon CloudFormation template [custom resource](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/template-custom-resources.html) function that runs when a resource with a `Custom` type that references the function is added, updated, or deleted.

### DIFF
--- a/package.json
+++ b/package.json
@@ -4,8 +4,8 @@
   "version": "0.0.1",
   "type": "module",
   "devDependencies": {
-        "@aws-sdk/client-athena": "^3.188.0",
-        "@aws-sdk/client-cloudwatch-logs": "^3.188.0",
+    "@aws-sdk/client-athena": "^3.188.0",
+    "@aws-sdk/client-cloudwatch-logs": "^3.188.0",
     "@aws-sdk/client-dynamodb": "^3.188.0",
     "@aws-sdk/client-s3": "^3.188.0",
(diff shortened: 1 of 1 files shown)


cloudformation/config-store.yaml
```diff
@@ -0,0 +1,46 @@
...
+            KMSMasterKeyID: !GetAtt KmsKey.Arn
+
+UkVatConfigS3Object:
+  Type: Custom::S3Object
+  Properties:
+    ServiceToken: !GetAtt CustomS3ObjectResourceFunction.Arn
+    S3Object:
+      Bucket: !Ref ConfigBucket
+      Key: uk-vat.json
+      Item: [{ "rate": 0.2, "start": "2011-01-04" }]
```
gareth-smyth-gov, line 46: Using our custom !YAMLInclude like `Item: !YAMLInclude ./cloudformation/uk-vat.json` seems to work here which would allow us to store these as files in source. I'm not sure how it would handle larger files on multiple lines.

gareth-smyth-gov: I think it's cleaner, especially once we add contract data etc. in the same way.

gareth-smyth-gov (approved):

PR #87, opened 2022-10-28
Fix tests

gdsgj:
This fix tests and linting errors and adds automatic unit test running on Git push

This also suppresses console logs in more tests so output is easier to read


gareth-smyth-gov (approved):

markschnitziusgovuk (approved):
lgtm

PR #91, opened 2022-11-01
Fix VAT bugs

gdsgj:
This fixes a bug where a number in the VAT JSON was stored as a string. It seems to have been caused by CloudFormation converting all YAML numbers to strings in custom-resource event properties without this being documented.

This also fixes a bug where CloudFormation automatically deleted updated custom resources. This was caused by using some of Amazon's example code that erroneously uses a new resource ID in custom-resource update results.


MelAtGDS (approved):

markschnitziusgovuk (approved):
Looks fine from what I could tell.

PR #105, opened 2022-11-11
Fix alarms

gdsgj:
This fixes a bug where the CloudWatch alarms were not being triggered by messages in dead letter queues. It was caused by using the wrong metric, due to a misunderstanding (`MessagesSent` are all messages sent *to* the queue, and `MessagesReceived` are messages counted *only after someone/something does a special "receive" action*)


markschnitziusgovuk (approved):
lgtm

PR #108, opened 2022-11-11
BTM-202: Add invoice storage infra

gdsgj:
This adds AWS resources for Textract invoice data storage. The Lambda function is just a placeholder that can be configured to return a value or throw an error depending on the environment variables for testing purposes

This also renames the Store function to "Store Transactions", because there is now another function whose main purpose is to store things

### DIFF
--- a/src/handlers/store-raw-invoice-textract-data/handler.ts
+++ b/src/handlers/store-raw-invoice-textract-data/handler.ts
@@ -1,4 +1,6 @@
-export const handler = async (): Promise<string | undefined> => {
+export const handler = async (event?: unknown): Promise<string | undefined> => {
+  console.log("event:", event);
+
   if (
     String(
       process.env.RAW_INVOICE_TEXTRACT_DATA_STORAGE_HANDLER_THROW_ERROR
--- a/src/handlers/extract/handler.test.ts
+++ b/src/handlers/extract/handler.test.ts
@@ -2,18 +2,15 @@ import { handler } from "./handler";
 
 const OLD_ENV = process.env;
 const oldConsoleError = console.error;
-const oldConsoleLog = console.log;
 
 beforeEach(() => {
   process.env = { ...OLD_ENV };
   console.error = jest.fn();
-  console.log = jest.fn();
 });
 
 afterAll(() => {
   process.env = OLD_ENV;
   console.error = oldConsoleError;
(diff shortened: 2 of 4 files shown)


cloudformation/invoice-storage.yaml
```diff
@@ -0,0 +1,9 @@
+InvoiceStorageQueue:
+  Type: AWS::SQS::Queue
+  Properties:
+    KmsMasterKeyId: !GetAtt KmsKey.Arn
+
+InvoiceStorageTopic:
```
MelAtGDS, line 6: I know this is still in draft but I thought I'd share something I came across which might impact this. 
Step 3 from the below says to prepend the SNS topic name with AmazonTextract.
https://docs.aws.amazon.com/textract/latest/dg/api-async-roles.html#api-async-roles-all-topics.

MelAtGDS: I need to refer to the SNS topic in my extraction function so I've gone with AmazonTextractSNSTopic. Is that ok?

MelAtGDS: Thanks Gareth, your reasoning makes sense. AmazonTextractRawInvoiceDataTopic it is!

cloudformation/raw-invoice-textract-data-store.yaml
```diff
@@ -1,3 +1,8 @@
+AmazonTextractRawInvoiceDataTopic:
```
MelAtGDS, line 1: Do you think we will need some kind of policy around this SNS topic? Like to say that it can invoke the RawInvoiceTextractDataQueue? 
@phinterseer what do you think?

src/handlers/store-raw-invoice-textract-data/handler.ts
```diff
@@ -0,0 +1,14 @@
+export const handler = async (): Promise<string | undefined> => {
```
MelAtGDS, line 1: Could we take in the event and console log it to make it easier for the next ticket that  builds out the function?

MelAtGDS: i see logging the event is in the extract handler (which I'm working on today). Did you mean to add the log to this handler?

MelAtGDS (approved):

PR #112, opened 2022-11-14
BTM-212 Start Textract Flow

MelAtGDS:


cloudformation/extraction.yaml
```diff
@@ -25,6 +26,14 @@ ExtractFunction:
               - sqs:GetQueueAttributes
               - sqs:ReceiveMessage
             Resource: !GetAtt ExtractQueue.Arn
+          - Effect: Allow
+            Action:
+              - "sns:*"
+            Resource: "*"
```
phinterseer, line 32: This should be more locked down to the actual resources.

phinterseer: Also, this is the function, it does not need SNS access as it is not doing anything with SNS.

cloudformation/extraction.yaml
```diff
@@ -25,6 +26,14 @@ ExtractFunction:
...
+          - Effect: Allow
+            Action:
+              - "textract:*"
+            Resource: "*"
```
phinterseer, line 36: Bot actions and resources can probably be locked down more. But it's OK to leave like this to get it running, but don't forget to tighten afterwards.

cloudformation/extraction.yaml
```diff
@@ -51,6 +55,35 @@ ExtractQueueDeadLetterQueue:
...
+      Statement:
+        - Effect: "Allow"
+          Principal:
+            Service:
+              - "textract.amazonaws.com"
+              - "lambda.amazonaws.com"
+          Action:
+            - "sts:AssumeRole"
+    ManagedPolicyArns:
+      - "arn:aws:iam::aws:policy/AWSLambdaExecute"
```
phinterseer, line 72: This should not be needed, as this is NOT the role that the lambda uses, it's one Textract assumes to write to the topic.

cloudformation/extraction.yaml
```diff
@@ -51,6 +55,35 @@ ExtractQueueDeadLetterQueue:
...
+          Version: "2012-10-17"
+          Statement:
+            - Effect: Allow
+              Action:
+                - "sns:*"
+              Resource: "*"
+            - Effect: Allow
+              Action:
+                - "textract:*"
+              Resource: "*"
```
phinterseer, line 85: This might not be needed as it's Textract assuming this role. Just test if it works without the last bit.

cloudformation/extraction.yaml
```diff
@@ -51,6 +55,35 @@ ExtractQueueDeadLetterQueue:
...
+      - "arn:aws:iam::aws:policy/AWSLambdaExecute"
+    Policies:
+      - PolicyName: "TextractRoleAccess"
+        PolicyDocument:
+          Version: "2012-10-17"
+          Statement:
+            - Effect: Allow
+              Action:
+                - "sns:*"
+              Resource: "*"
```
phinterseer, line 81: This should be locked down to the actual topic as well. Actions can very likely be limited as well.

src/handlers/extract/handler.ts
```diff
@@ -1,12 +1,47 @@
...
+  const response: AWS.Textract.StartExpenseAnalysisResponse[] = [];
+
+  // Get Bucket and filename from the event.
+  const promises = event.Records.map(async (record) => {
+    console.log(`Record ${JSON.stringify(record)}`);
+    const bucket = record.s3.bucket.name;
+    const fileName = record.s3.object.key;
+    console.log("File Name: " + fileName);
+
+    if (process.env.TEXT_EXTRACT_ROLE === undefined) {
```
phinterseer, line 18: You should check this and the SNS-topic config first thing in the function as you will not be able to continue without it anyway.

src/handlers/extract/handler.ts
```diff
@@ -1,12 +1,47 @@
...
-    console.error(message);
-    throw new Error(message);
-  }
-
-  return process.env.EXTRACTION_HANDLER_RETURN_VALUE;
+import * as AWS from "aws-sdk";
+import { S3Event } from "aws-lambda";
+
+export const handler = async (
+  event: S3Event
```
phinterseer, line 5: This will not be of type S3event. It will be an SQS event.

src/handlers/extract/handler.ts
```diff
@@ -1,12 +1,47 @@
...
+        },
+      },
+      NotificationChannel: {
+        RoleArn: process.env.TEXT_EXTRACT_ROLE,
+        SNSTopicArn: process.env.SNS_TOPIC,
+      },
+    };
+
+    // Invoke textract function
+    response.push(await textract.startExpenseAnalysis(params).promise());
```
phinterseer, line 42: Better just return the promise, let the `.map` do the collecting and get rid of that whole mutable array `response`.

src/handlers/extract/handler.ts
```diff
@@ -1,12 +1,47 @@
...
+
+export const handler = async (
+  event: S3Event
+): Promise<AWS.Textract.StartExpenseAnalysisResponse[]> => {
+  // Set up
+  const textract = new AWS.Textract({ region: "eu-west-2" });
+  const response: AWS.Textract.StartExpenseAnalysisResponse[] = [];
+
+  // Get Bucket and filename from the event.
+  const promises = event.Records.map(async (record) => {
```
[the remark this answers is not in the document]
phinterseer: Mel showed me that batch failure responses are implemented in the other lambdas, so this should be all good as soon as we implement those here as well.

src/handlers/extract/handler.ts
MelAtGDS: Pending incoming event...

src/handlers/extract/handler.ts
MelAtGDS: Pending incoming event...

cloudformation/extraction.yaml
```diff
@@ -48,9 +57,39 @@ Resources:
...
+                Action:
+                  - "s3:GetObject*"
+                Resource: !GetAtt RawInvoicePdfBucket.Arn
+              - Effect: Allow
+                Action:
+                  - "sns:Publish"
+                Resource: !Ref AmazonTextractRawInvoiceDataTopic
+              - Effect: Allow
+                Action:
+                  - kms:*
```
phinterseer, line 90: Better lock this down a little as this also opens up key-management like kms:DisableKey etc.

src/handlers/extract/handler.test.ts
```diff
@@ -1,41 +1,277 @@
...
+    };
+
+    const result = await handler(givenEvent as SQSEvent);
+
+    expect(result).toEqual({
+      batchItemFailures: [{ itemIdentifier: "given message ID" }],
+    });
+  });
+
+  test("Extract handler with event record that has body that does not serialises to object", async () => {
```
phinterseer, line 149: serialise

src/handlers/extract/handler.ts
```diff
@@ -1,14 +1,78 @@
...
+import { S3Event, SQSEvent } from "aws-lambda";
+import { Response } from "../../shared/types";
+
+export const handler = async (event: SQSEvent): Promise<Response> => {
+  // Set Up
+  const textExtractRole = process.env.TEXTRACT_ROLE;
+  const snsTopic = process.env.TEXTRACT_SNS_TOPIC;
+
+  if (textExtractRole === undefined || textExtractRole === "") {
+    const textractMessage = "Textract role not set.";
```
phinterseer, line 11: I think you can just inline the variable in the throw. The variable name is not adding much.

src/handlers/extract/handler.ts
```diff
@@ -1,14 +1,78 @@
...
+    throw new Error(textractMessage);
+  }
+  if (snsTopic === undefined || snsTopic === "") {
+    const snsMessage = "SNS Topic not set.";
```
phinterseer, line 15: Same as with the other throw.

src/handlers/extract/handler.ts
```diff
@@ -1,14 +1,78 @@
...
+      console.log(e);
+      response.batchItemFailures.push({ itemIdentifier: record.messageId });
+    }
+  });
+
+  await Promise.all(promises);
+  return response;
 };
+
+const isS3Event = (object: any): object is S3Event =>
```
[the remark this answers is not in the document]
MelAtGDS: This was Gareth's doing :)

phinterseer (dismissed):

cloudformation/extraction.yaml
MelAtGDS: Putting this in and then will deploy to check the textract-sns-sqs flow

PR #114, opened 2022-11-16
BTM-184: Fix alarms

gdsgj:
This fixes a bug where queue alarms were not being triggered. It was caused by a misunderstanding of the `NumberOfMessagesSent` metric (it is the number of messages sent to *any* queue *by the given queue*, not messages sent *to* the given queue from anywhere)


hamish-pollock-gds (approved):
Reviewed and approved - to unblock @NKannan14 for testing

PR #118, opened 2022-11-21
Fix deployment & add log

gdsgj:


MelAtGDS:
@gdsgj Can you please add a log to the extract event so we can see what it looks like? I opened a PR for it here but it might make sense just to add it to your deployment fix PR.
https://github.com/alphagov/di-billing-transaction-monitoring/pull/119

MelAtGDS (approved):

PR #120, opened 2022-11-21
Chore/dynamic names for cf

phinterseer:
The CF files all changed as I added the CF header to all of them and put everything under `Resources` to allow for IntelliJ filetype detection to work. Changes to the CF code are mainly about names and removing unnecessary DLQs.

A few integration tests are still failing here, but I open this up for comments while I fix this.
UPDATE: Not actually true, I just needed to rebuild and deploy after rebase... The athena-test is failing, though, but I think there is something missing in the infra to make it work. The data in the bucket does not seem to be crawled into a DB.

Also please give it a go on your machines to see if everything works.

Thanks.

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,7 +12,7 @@ If you are new to the Billing & Transaction Monitoring team, please also read [o
 
 ## Check before commit
 ```sh
-npm run build-template
+npm run build:template
 checkov -f template.yaml --framework cloudformation --external-checks-git git@github.com:alphagov/di-devplatform-checkov-hook.git//src/gds_digitalidentity_checkovhook/custom_policies
 ```
 N.B. You may get Python errors due to conflicting dependencies with SAM CLI. If you do, run this and try again:
@@ -47,7 +47,7 @@ docker run --rm -it -p 4566:4566 -p 4571:4571 -e LOCALSTACK_DEBUG=1 localstack/l
 ### Bringing up the stack
 ```sh
 npm run build
-npm run build-template NO_LOCAL
+npm run build:template NO_LOCAL
 samlocal build
 samlocal deploy --resolve-s3 --config-env local
 ```
@@ -90,7 +90,7 @@ npm run test:integration
 To run the tests against local environment
 ````
 npm i
-npm run test:integration-local
+npm run test:integration:local
 ````
 
 To generate  emailable allure report after running the integration test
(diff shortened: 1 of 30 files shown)


NKannan14:
Hi @phinterseer yes, Athena test is failing in the main branch as well, BTM-227 crawler setup needs to be done . Integration tests changes looks fine, I will run it my machine and see

README.md
```diff
@@ -108,7 +108,26 @@ npm run beforeIntTest
 
 ## Deploy
 
-To deploy to the `dev` environment (also called `build`), commit to the `main` branch
+To deploy to the `dev` environment on the main stack `di-btm`, commit to the `main` branch, this will also deploy into
+the `build` environment in the build account.
+
+To deploy to your own on-demand stack:
+
+Log in to the `dev` account using your method of choice (`gds-cli`, AWS profiles, etc.).
```
MelAtGDS, line 116: I think that, because of the information in the new starter guide, we log in to dev using the gds-cli like this: gds aws di-btm-dev -s
Do we replace dev with build or prod to log into the other environments?

cloudformation/athena.yaml
```diff
@@ -1,99 +1,104 @@
...
-        SerdeInfo:
-          Parameters: { "ignore.malformed.json": true,"serialiazation.format": 1,"field.delim": "" }
-          SerializationLibrary: "org.openx.data.jsonserde.JsonSerDe"
+Resources:
+  TransactionDB:
+    Type: AWS::Glue::Database
+    Properties:
+      CatalogId: !Ref AWS::AccountId
+      DatabaseInput:
+        Name: !Join [ '-', [!Ref Prefix, !Ref Environment, transactions ] ]
```
MelAtGDS, line 9: What is the Prefix? !Ref Prefix means that it will return the value of Prefix right? Where is that declared?

cloudformation/athena.yaml
```diff
@@ -1,99 +1,104 @@
...
-          EncryptionOption: "SSE_KMS"
-          KmsKey: !GetAtt KmsKey.Arn
-        OutputLocation: !Join [ '', ['s3://', !Ref AthenaQueryResultsBucket, '/']]
+  TransactionTable:
+    Type: AWS::Glue::Table
+    Properties:
+      CatalogId: !Ref AWS::AccountId
+      DatabaseName: !Ref TransactionDB
+      TableInput:
+        Name: btm_transactions
```
MelAtGDS, line 17: @markschnitziusgovuk Should the Name of the Glue Table be the same as the Glue Database above?

README.md
```diff
@@ -108,7 +108,26 @@ npm run beforeIntTest
...
+Then run:
+
+```
+export ENV_NAME=dev-od-ph-purpose
+```
+Replace `ph` with your initials and replace `something` with a very short purpose of the env like `testing`.
```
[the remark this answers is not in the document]
phinterseer: I did, will correct.

cloudformation/athena.yaml
phinterseer: Prefix is one of the parameters this stack has. They are defined in `template-source.yaml`. This is to be easily able to rename everything in case we want to split into multiple stacks at some point.

phinterseer:
> How does this affect the alert-chatbot-template.yaml? We have these three slack channels: #di-btm-errors-dev #di-btm-errors-staging #di-btm-errors

I have not explicitly looked into this, but from a cursory glance it should be OK to use, just use the correct name of the SnsTopicArn when deploying it.

integration_tests/helpers/envHelper.ts
```diff
@@ -0,0 +1,5 @@
+function resourcePrefix() {
+  return `di-btm${process.env.ENV_NAME ? '-' + process.env.ENV_NAME : ''}`;
```
markschnitziusgovuk, line 2: Shouldn't the `di-btm` prefix part of this be pulled in from config as well?

markschnitziusgovuk (approved):

integration_tests/helpers/envHelper.ts
phinterseer: There is currently nowhere I could pull it from. If we were to create another project to deploy separately, it would probably have it's own set of tests, so this would be hard-coded as something else there.

PR #124, opened 2022-11-24
BTM-213: Add raw invoice Textract data storing

gdsgj:
This adds automatic storage of raw invoice Textract data

### DIFF
--- a/.prettierrc
+++ b/.prettierrc
@@ -1,0 +1,7 @@
+overrides:
+  - files:
+      - '*.yaml'
+      - '*.yml'
+      - .prettierrc
+    options:
+      singleQuote: true
--- a/README.md
+++ b/README.md
@@ -133,7 +133,9 @@ To run the texract function integration test
 
 The test invoice file needs to be manually uploaded into test S3 bucket (ENV_NAME-test-invoice-pdf) with name Invoice.pdf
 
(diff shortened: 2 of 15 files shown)


src/handlers/store-raw-invoice-textract-data/store-expense-documents.ts
```diff
@@ -0,0 +1,43 @@
...
+): Promise<void> {
+  const { jobId, sourceBucket, sourceFileName, status } =
+    getQueuedExpenseAnalysisNotificationData(record);
+
+  const [sourceFolder] = sourceFileName.split("/");
+  if (
+    sourceFolder === RAW_INVOICE_TEXTRACT_DATA_FOLDER_FAILURE ||
+    sourceFolder === RAW_INVOICE_TEXTRACT_DATA_FOLDER_SUCCESS
+  )
+    return;
```
markschnitziusgovuk, line 23: I think maybe I don't understand what's meant to happen here, but it's not clear to me why we would just return no value in this case, and not log anything.

markschnitziusgovuk: Either of those would be fine, or else consider pulling the whole section of code into a helper method, e.g. alreadyProcessed().

markschnitziusgovuk (approved):

PR #125, opened 2022-11-24
Btm 227 add start crawler

NKannan14:


NKannan14:
closing this PR as no need to start the crawler before tests

PR #128, opened 2022-11-28
Remove storage to dynamoDB

MelAtGDS:


markschnitziusgovuk (dismissed):

cloudformation/transaction-storage.yaml
```diff
@@ -1,11 +1,11 @@
-AWSTemplateFormatVersion: '2010-09-09'
+AWSTemplateFormatVersion: "2010-09-09"
 Transform: AWS::Serverless-2016-10-31
 
 Resources:
   StorageBucket:
     Type: AWS::S3::Bucket
     Properties:
-      BucketName: !Join [ '-', [!Ref Prefix, !Ref Environment, storage ] ]
+      BucketName: !Join ["-", [!Ref Prefix, !Ref Environment, storage]]
```
phinterseer, line 8: I like the spaces around the brackets, makes this horrible CF-syntax a bit more readable.

phinterseer (dismissed):
Didn't check if this was everything DynamoDB related, but it looks OK.

cloudformation/transaction-storage.yaml
MelAtGDS: the linting is auto doing this on save :(

PR #150, opened 2022-12-08
BTM-311 invoice Athena & Glue

gdsgj:


src/handlers/custom-athena-view-resource/handler.ts
```diff
@@ -0,0 +1,122 @@
...
+interface AthenaViewResourceProperty {
+  Database: string;
+  Name: string;
+  Query: string;
+  Workgroup: string;
+}
+
+const QUERY_EXECUTION_RETRIEVAL_INTERVAL_MS = 1000;
+const QUERY_EXECUTION_RETRIEVAL_MAX_ATTEMPTS = 10;
+
```
markschnitziusgovuk, line 14: I think Mel's PR includes this same file -- I don't know whose will go in first, but just be sure to share any modifications you do to it.

I commented in hers that this function seems a bit long as well, and wondered if maybe the initial validation can be pulled into a separate function, or if there could be a helper class around the query retrying maybe.

MelAtGDS: Thanks Mark. The PR for 228-transaction view will be closed as the ticket was closed and not needed.

markschnitziusgovuk (approved):

PR #152, opened 2022-12-08
BTM-307: Run integration tests in pipeline

phinterseer:
Still running unit tests in pipeline, but it is good to review now. Should not be much more to run the integration tests instead.


integration_tests/tests/sns-s3-tests.ts
```diff
@@ -1,67 +1,69 @@
...
+import { waitForTrue } from "../../src/handlers/int-test-support/helpers/commonHelpers";
+import { publishSNS } from "../../src/handlers/int-test-support/helpers/snsHelper";
 
 let snsResponse: PublishResponse;
 const storageBucket: string = `${resourcePrefix()}-storage`;
 const objectsPrefix = "btm_transactions";
 
 console.log(storageBucket);
 
+const checkForEventId = async (
```
markschnitziusgovuk, line 25: Can this method have a more specific name?  Check where?

src/handlers/int-test-support/helpers/snsHelper.ts
```diff
@@ -0,0 +1,35 @@
...
+let snsParams: PublishInput;
+
+const snsParameters = async (
+  snsValidEventPayload: any
+): Promise<{
+  Message: string;
+  TopicArn: string;
+}> => ({
+  Message: JSON.stringify(snsValidEventPayload),
+  TopicArn: `arn:aws:sns:eu-west-2:582874090139:${resourcePrefix()}-test-TxMA-topic`,
```
markschnitziusgovuk, line 19: I know this is fixed and unlikely to change, but it still seems wrong to have the acct # hardcoded in a public repo.  Should this be read from config?

integration_tests/tests/billing-and-transaction-view-tests.ts
```diff
@@ -29,65 +32,65 @@ describe("\nUpload invoice to standardised folder and verify billing and transac
...
+    await deleteS3Objects({ bucketName, prefix: "btm_transactions" });
     // uploading file to s3 will be removed once BTM-276 changes merged
     const file = "../payloads/receipt.txt";
     const filePath = path.join(__dirname, file);
-    const fileStream = fs.createReadStream(filePath);
-    await putObjectToS3(testObject, fileStream);
+    const fileData = fs.readFileSync(filePath);
+    await putS3Object({ target: testObject, data: fileData });
     const checkFileExists = await checkIfS3ObjectExists(testObject);
     expect(checkFileExists).toBe(true);
```
Si3rr4wow, line 44: I know this isn't part of the diff and would be one to address in another (smaller) PR but how do we feel about using `expect` like this?

integration_tests/tests/s3-invoice-raw-store-tests.ts
```diff
@@ -47,7 +39,7 @@ describe("\n Happy path S3 raw-invoice-pdf and raw-invoice-textract-data bucket
           (item) => {
             return (
               item.LastModified !== undefined &&
-              item.LastModified >= testStartTime
+              new Date(item.LastModified) >= testStartTime
```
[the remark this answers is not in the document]
Si3rr4wow: won't it be a timestamp as a string since this is an API response?

src/handlers/int-test-support/handler.test.ts
```diff
@@ -0,0 +1,71 @@
...
+  console.error = jest.fn();
+
+  givenContext = {
+    someContextItem: "some context item",
+  } as any;
+});
+
+afterAll(() => {
+  console.error = oldConsoleError;
+});
```
Si3rr4wow, line 29: Is it ok to do these things at top level? I've only ever seen them nested in `describes`, will they be scoped to this file?

src/handlers/int-test-support/handler.ts
```diff
@@ -0,0 +1,70 @@
...
+import { createInvoiceInS3 } from "./helpers/mock-data/invoice/helpers";
+
+export interface TestSupportEvent {
+  environment: string;
+  command: string;
+  parameters: any;
+}
+
+export interface TestSupportReturn {
+  error?: string;
```
Si3rr4wow, line 27: I don't think this gets used

src/handlers/int-test-support/handler.ts
```diff
@@ -0,0 +1,70 @@
...
+  success?: boolean;
+  successObject?: any;
+}
+
+const functionMap: { [name: string]: Function } = {
```
Si3rr4wow, line 32: probably fine in this case cus it's for test tooling but I'd normally discourage the use of `Function` as a type

integration_tests/Dockerfile
```diff
@@ -0,0 +1,59 @@
...
+      build-base \
+      cmake \
+      git \
+      groff \
+      libffi-dev
+RUN git clone --single-branch --depth 1 -b ${AWS_CLI_VERSION} https://github.com/aws/aws-cli.git
+
+WORKDIR /aws-cli
+
+RUN sed -i'' 's/PyInstaller.*/PyInstaller==5.6.2/g' requirements-build.txt
```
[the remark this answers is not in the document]
phinterseer: That bit is copied from our other team. I didn't even see this, does not make much sense to me. I'll try to remove it.

integration_tests/Dockerfile
```diff
@@ -0,0 +1,59 @@
...
+RUN apk update && \
+      apk upgrade && \
+      apk add --no-cache \
+      bash
+
+WORKDIR /test-app
+
+COPY package.json package-lock.json tsconfig.json jest.integration.config.ts ./
+COPY integration_tests ./integration_tests
+COPY src/handlers/int-test-support/ ./src
```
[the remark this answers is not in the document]
phinterseer: The bit later is for running the unit-tests, which will go away again once this container runs the integration tests as it is supposed to.

integration_tests/run-tests.sh
```diff
@@ -0,0 +1,22 @@
+#!/bin/bash
+
+# This script will only run in AWS Codepipeline. It has access to the following environment variables:
+# CFN_<OUTPUT-NAME> - Stack output value (replace <OUTPUT-NAME> with the name of the output)
+# TEST_REPORT_ABSOLUTE_DIR - Absolute path to where the test report file should be placed
+# TEST_REPORT_DIR - Relative path from current directory to where the test report file should be placed
+# TEST_ENVIRONMENT - The environment the pipeline is running the tests in
+
+# This file needs to be located at the root when running in the container. The path /tests is defined
+# in the Dockerfile.
```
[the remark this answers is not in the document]
phinterseer: Also copied from TxMA and not questioned further. Will adapt.

integration_tests/tests/s3-invoice-raw-store-tests.ts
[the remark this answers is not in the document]
phinterseer: Good spot and good questions. :-) This is necessary for the case when the AWS calls run through the lambda. The parameters for the functions and the return values are sent as pure JSON payload and therefore reach the function as a date-string, which needs to be turned into a date. In the case where it does no run through the lambda, it already is a date, but the code needs to run in both cases, so this is the way to make this work.

integration_tests/tests/billing-and-transaction-view-tests.ts
phinterseer: Well, I found many things I didn't like in the code I worked on and I could not fix everything. This is one of those things that I saw but didn't see as it needs fixing immediately, but I do agree.

src/handlers/int-test-support/handler.test.ts
phinterseer: Yeah, I think they are, but I would rather have them in describes as well. Not sure if that's the scope of my work here though.

src/handlers/int-test-support/handler.ts
phinterseer: Well spotted, I don't think it is. :-)

src/handlers/int-test-support/handler.ts
phinterseer: I agree with you fully, but this would be a huge mess of code without it.

src/handlers/int-test-support/helpers/athenaHelper.ts
```diff
@@ -0,0 +1,124 @@
...
+  };
+  return await athenaClient.send(new GetQueryResultsCommand(params));
+};
+
+const waitAndGetQueryResults = async (
+  queryId: string
+): Promise<GetQueryResultsCommandOutput | undefined> => {
+  const checkState = async (): Promise<boolean> => {
+    const result = await getQueryExecutionStatus(queryId);
+    return !(result?.State?.match("SUCCEEDED") == null);
```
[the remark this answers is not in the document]
phinterseer: This is what IntelliJ + eslint make of it when fixing the nullability issue... I would not write it like that as it's horrible to read, but it's the only way I found that the linter and code inspection accepts it.

src/handlers/int-test-support/helpers/athenaHelper.ts
```diff
@@ -0,0 +1,124 @@
...
+  const formattedData = rows.map((row) => {
+    const object: StringObject = {};
+    row?.forEach(function (_, index) {
+      const fieldName = columns[index].VarCharValue;
+      const fieldValue = row[index].VarCharValue;
+      if (fieldName !== undefined && fieldValue !== undefined) {
+        object[fieldName] = fieldValue;
+        return object;
+      }
+    });
```
Si3rr4wow, line 104: Bit more stylish to use reduce for this
```suggestion
    row.reduce<Record<string, string>>((accj, _, index) => {
      const fieldName = columns[index].VarCharValue;
      const fieldValue = row[index].VarCharValue;
      if (fieldName === undefined || fieldValue === undefined) {
        return accj;
      }
      return {
        ...accj,
        [fieldName]: fieldValue,
      };
    }, {})
```

src/handlers/int-test-support/helpers/snsHelper.ts
phinterseer: This also means that this test will not work in the build account (where it should be able to run), so I'll have to fix this somehow...

Si3rr4wow (approved):

integration_tests/tests/sns-s3-tests.ts
phinterseer: renamed

src/handlers/int-test-support/helpers/athenaHelper.ts
phinterseer: Yes, much nicer and also non-mutating. Will use that.

src/handlers/int-test-support/helpers/snsHelper.ts
phinterseer: Added node-cache and look this ARN in run-time now (once).

PR #164, opened 2022-12-20
fixed the s3 objects map

NKannan14:
fixed hard coded receipt invoice input file decimal precisions
fixed mapping of s3Objects instead query


integration_tests/tests/invoice-athena-tests.ts
```diff
@@ -38,7 +38,7 @@ describe("\nExecute athena query to retrive invoice data and validate it matches
       (val) => strFromQuery[val]
     );
     const s3Array = await s3GetObjectsToArray(bucketName, folderPrefix);
-    const s3Objects = Object.keys(s3Array).map((val) => strFromQuery[val]);
+    const s3Objects = Object.keys(s3Array).map((val) => s3Array[val]);
```
Si3rr4wow, line 41: If I'm reading this right I think you just want `Object.values(s3Array)`

markschnitziusgovuk (approved):

integration_tests/tests/invoice-athena-tests.ts
Si3rr4wow: perf!

PR #166, opened 2022-12-22
Btm 319 fake pdf

Si3rr4wow:


cloudformation/fake-prices.csv
```diff
@@ -1,9 +1,10 @@
...
\ No newline at end of file
+client1,Passport check,0,,1.00,2022-01-01
+client1,Fraud check,0,,1.505,2022-01-01
+client2,Passport check,0,,2.00,2022-01-01
+client2,Kbv check,0,,3.99,2022-01-01
+client3,Passport check,0,10,4.00,2022-01-01
+client3,Passport check,11,,0.50,2022-01-01
+client3,Address check,0,5,6.50,2022-01-01
+client3,Address check,6,10,0.25,2022-01-01
+client3,Address check,11,,8.88,2022-01-01
```
markschnitziusgovuk, line 10: This file has changed on the 241 branch; it now contains `event_name`s instead of `service_name`s.  I suggest you merge that branch in to yours (or main if you read this after that branch merges).

markschnitziusgovuk (changes requested):

cloudformation/fake-prices.csv
Si3rr4wow: yeah looks like there's a fair whack of conflict resolution to do here after recent changes 😅

MelAtGDS: mm and this file is deleted from peter's PR #168. I think I'm a bit confused about where these fake files have gone...

Si3rr4wow: This diff is kerblamo

markschnitziusgovuk (approved):

PR #167, opened 2022-12-22
Btm 241 int tests

NKannan14:


integration_tests/helpers/commonHelpers.ts
```diff
@@ -1,3 +1,12 @@
+import { generateRandomNumber, snsValidEventPayload } from "../payloads/snsEventPayload";
+import { deleteObjectInS3, getS3ItemsList } from "./s3Helper";
+import { publishSNS } from "./snsHelper";
+import { resourcePrefix } from "../helpers/envHelper";
+
+const prefix = resourcePrefix();
+const objectsPrefix = "btm_transactions";
+const details: string[] = [];
```
Si3rr4wow, line 8: I think it would be cleaner to move this declaration into the `generateTestEventsAndValidateEventExists` function then change `deleteS3Event` to take an array of eventIds as an argument

integration_tests/helpers/commonHelpers.ts
```diff
@@ -19,5 +28,47 @@ async function waitForTrue(
...
+    const checkEventId = async (): Promise<boolean> => {
+      const result = await getS3ItemsList(`${prefix}-storage`, objectsPrefix);
+      if (result.Contents !== undefined) {
+        return JSON.stringify(result.Contents.map((data) => data.Key)).includes(
+          snsValidEventPayload.event_id
+        );
+      } else {
+        console.log("Storage bucket contents empty");
+        return false;
+      }
```
Si3rr4wow, line 52:
```suggestion
    if (result.Contents === undefined) {
      console.log("Storage bucket contents empty");
      return false;
    }
    return JSON.stringify(result.Contents.map((data) => data.Key)).includes(
      snsValidEventPayload.event_id
    );
```

This is another one of those denestable statements I mentioned yesterday

integration_tests/payloads/snsEventPayload.ts
```diff
@@ -93,5 +93,5 @@ export {
   snsEventMissingTimestampPayload,
   snsEventWithAdditionalFieldsPayload,
   snsMissingEventNamePayload,
-  snsEventMisisingEventIdValue,
+  snsEventMisisingEventIdValue,generateRandomNumber
```
Si3rr4wow, line 96: Top of  out January task list has to be getting your linter playing nice with your editor 😅

integration_tests/tests/transaction-view-athena-tests.ts
```diff
@@ -0,0 +1,73 @@
...
+    await generateTestEventsAndValidateEventExists(
+      14,
+      "IPV_ADDRESS_CRI_END",
+      "client3"
+    );
+    const response = await queryResults();
+    expect(expectedCalculatedPrice).toEqual(response[0].price);
+  });
+
+  test.only("no results returned from billing_curated athena view query when the event payload has invalid eventName", async () => {
```
Si3rr4wow, line 53: did you mean to leave this `only` in?

integration_tests/tests/transaction-view-athena-tests.ts
```diff
@@ -0,0 +1,73 @@
...
+  waitForTrue,
+} from "../helpers/commonHelpers";
+import { publishSNS } from "../helpers/snsHelper";
+import { snsInvalidEventNamePayload } from "../payloads/snsEventPayload";
+
+const prefix = resourcePrefix();
+const databaseName = `${prefix}-calculations`;
+
+describe("\nExecute athena query to retrive transaction data\n", () => {
+  test("price retrived from billing_curated athena view query should match with expected calculated price for 2 events", async () => {
```
Si3rr4wow, line 20: Not essential but you could rework these 3 cases using `test.each`, here's an article about it that I reference whenever I need the syntax https://elfi-y.medium.com/reduce-boilerplate-test-code-with-jest-it-each-30a0eec9776d

integration_tests/helpers/commonHelpers.ts
NKannan14: By moving the const details to inside are you suggesting to change the return type of generateTestEventsAndValidateEventExists function and return details ?

integration_tests/helpers/commonHelpers.ts
NKannan14: yes thanks will change it

integration_tests/payloads/snsEventPayload.ts
NKannan14: yes need to make ESLint to work in Editor

integration_tests/tests/transaction-view-athena-tests.ts
NKannan14: will remove this only

integration_tests/tests/transaction-view-athena-tests.ts
NKannan14: looks this is like Example (data table ) in cucumber if not essential may be one for future tests or can do part of refactoring

integration_tests/helpers/commonHelpers.ts
Si3rr4wow: Yeah I'd do something like that. The way you've done it will work fine but it is prone to bugs if some other function is introduced later that also depends on the details array

MelAtGDS (approved):

integration_tests/helpers/commonHelpers.ts
```diff
@@ -1,23 +1,103 @@
...
+  console.log("deleted the file from s3");
+  return true;
+};
 
-export { waitForTrue };
+export const deleteS3Events = async (eventIds: string[]): Promise<boolean> => {
+  for (const eventId of eventIds) {
+    await deleteS3Event(eventId);
+  }
+  console.log("deleted the file from s3");
```
Si3rr4wow, line 101:
```suggestion
  console.log("deleted the files from s3");
```

Si3rr4wow (approved):

integration_tests/helpers/commonHelpers.ts
NKannan14: fixed

PR #168, opened 2022-12-22
Use the separate config stack di-btm-cfg-$ENV.

phinterseer:
Removed parts that went into the config stack. Also some test refactoring.

Once this is merged:

I recommend deleting your ephemeral stacks before applying this change or there will be a lost config bucket flying around that you need to delete manually (which is also the alternative if you don't heed this advice :-) ).


NKannan14:
@phinterseer Integration test changes looks good to me

integration_tests/helpers/envHelper.ts
```diff
@@ -1,7 +1,11 @@
...
   }`;
-}
 
-export { resourcePrefix };
+const configStackName = (): string =>
+  `di-btm-config-${
+    process.env.CONFIG_NAME !== undefined ? process.env.CONFIG_NAME : "dev"
+  }`;
+
+export { resourcePrefix, configStackName };
```
Si3rr4wow, line 11: nit pick but why not just `export` each `const`?

phinterseer: I'd prefer that as well, but that's how it's done across the codebase, so I didn't mess with it.

Si3rr4wow (approved):
LGTM

integration_tests/helpers/envHelper.ts
Si3rr4wow: okily-dokily

markschnitziusgovuk (approved):
Yep, all looks good.

Si3rr4wow (approved):
👍

PR #172, opened 2023-01-05
BTM-320: Fix for failures in dev environment

phinterseer:
This looks far more straight forward than it was to find this solution. Somewhere on the way were about 10 nested cloudformation functions to extract and combine the correct parts of the stacks, but luckily I could resolve that...

TIL: If you have to use a complicated string interpolation in multiple places, put it in an SSM-Parameter and reference that instead (since there are no variables in CF...).

### DIFF
--- a/package.json
+++ b/package.json
@@ -61,7 +61,7 @@
     "build:template": "node ./src/tools/yaml-import.js",
     "checkov": "checkov --quiet -f template.yaml --framework cloudformation --external-checks-git git@github.com:alphagov/di-devplatform-checkov-hook.git//src/gds_digitalidentity_checkovhook/custom_policies",
     "sam:build": "npm run build && npm run build:template && npm run checkov && sam build",
-    "sam:deploy": "sam deploy --stack-name di-btm-${ENV_NAME} --parameter-overrides Environment=${CONFIG_NAME:-dev} --s3-prefix ${ENV_NAME}",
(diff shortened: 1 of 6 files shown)


Si3rr4wow (approved):

package.json
```diff
@@ -61,7 +61,7 @@
     "build:template": "node ./src/tools/yaml-import.js",
     "checkov": "checkov --quiet -f template.yaml --framework cloudformation --external-checks-git git@github.com:alphagov/di-devplatform-checkov-hook.git//src/gds_digitalidentity_checkovhook/custom_policies",
     "sam:build": "npm run build && npm run build:template && npm run checkov && sam build",
-    "sam:deploy": "sam deploy --stack-name di-btm-${ENV_NAME} --parameter-overrides Environment=${CONFIG_NAME:-dev} --s3-prefix ${ENV_NAME}",
+    "sam:deploy": "sam deploy --stack-name di-btm-${ENV_NAME} --parameter-overrides Environment=dev PrivateConfigStack=${CONFIG_NAME:-none} --s3-prefix ${ENV_NAME}",
```
MelAtGDS, line 64: Will the README need to be updated as well to explain how to deploy config stacks now?

phinterseer: The mechanism for the user stays the same, it should only require the CONFIG_NAME env var to be set, but I'll take another look and clarify if necessary.

MelAtGDS (approved):

PR #174, opened 2023-01-06
Btm 241 additional test

NKannan14:
added additional test for event generated  last year to check the effective from price

### DIFF
--- a/integration_tests/helpers/commonHelpers.ts
+++ b/integration_tests/helpers/commonHelpers.ts
@@ -22,9 +22,14 @@ export const thisTimeLastYear = (): number => {
   return Math.floor((new Date().getTime() - 365 * 24 * 60 * 60 * 1000) / 1000);
 };
 
+export enum TimeStamps {
+  THIS_TIME_LAST_YEAR,
+  CURRENT_TIME,
+}
+
 export const eventTimeStamp = {
-  THIS_TIME_LASTYEAR: thisTimeLastYear(),
-  CURRENT_TIME: validTimestamp(),
+  [TimeStamps.THIS_TIME_LAST_YEAR]: thisTimeLastYear(),
+  [TimeStamps.CURRENT_TIME]: validTimestamp(),
 };
 
 export const waitForTrue = async (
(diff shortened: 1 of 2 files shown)


integration_tests/helpers/commonHelpers.ts
```diff
@@ -42,13 +51,15 @@ export const waitForTrue = async (
...
+ ): Promise<SNSEventPayload> => {
+  const key = eventTime;
   return {
     event_name: eventName,
     event_id: generateRandomId(),
     component_id: "TEST_COMP",
-    timestamp: validTimestamp(),
+    timestamp:eventTimeStamp[key as keyof typeof eventTimeStamp],
     client_id: clientId,
   };
```
Si3rr4wow, line 64: Now that we're adding to the list of parameters we can pass in to this generated event it might be good to rework the interface on this function a bit.

```
export const generateTestEvent = async (
  overrides: Partial<SNSEventPayload> &
    Pick<SNSEventPayload, "event_name" | "client_id">
): Promise<SNSEventPayload> => ({
  event_id: generateRandomId(),
  component_id: "TEST_COMP",
  timestamp: validTimestamp(),
  ...overrides,
});

...

await generateTestEvent({
  client_id: ClientId.client1,
  event_name: EventName.IPV_ADDRESS_CRI_END,
  timestamp: thisTimeLastYear(),
  event_id: 'My_new_id'
});
```

This structure still mandates that we supply a clientId and eventName _and_ it lets us overwrite any other fields we might want to. It also used the `validTimestamp` function by default so we don't need to supply one unless we explicitly want to override it.

NKannan14: changed the generateTestEvent() interface with the suggested change. But for await generateTestEvent() is still okay to use timestamp value as  "timestamp: eventTimeStamp[eventTime as keyof typeof eventTimeStamp]" as the timestamp is given as input in jest table

MelAtGDS (approved):

Si3rr4wow (approved):

PR #177, opened 2023-01-10
Btm 233 tests

NKannan14:
1. Added integration tests for billing and transaction curated view
2. refactored queryResults function in transaction-view-tests and moved to helper class and named queryResponseFilterByVendorServiceNames so that both billing  and transaction view tests can use.
3. In deleteDirectoryRecursiveInS3 method in s3 helper throwing when running the tests individually so refactored the condition to return empty []
4.updated receipt.txt and removed the unused variables
5. made the test to run sequentially to avoid failures in billing and transaction view tests


integration_tests/helpers/commonHelpers.ts
```diff
@@ -2,7 +2,7 @@ import {
   ClientId,
   SNSEventPayload,
   EventName,
-} from "../payloads/snsEventPayload";
+ } from "../payloads/snsEventPayload";
```
markschnitziusgovuk, line 5: (minor) remove extra space

integration_tests/tests/billing-and-transaction-view-tests.ts
```diff
@@ -0,0 +1,176 @@
...
+    const checkFileExists = await checkIfS3ObjectExists(testObject);
+    expect(checkFileExists).toBeTruthy();
+  });
+
+  test.each`
+    eventName                          | clientId     | eventTime                  | numberOfTestEvents | priceDiff   | qtyDiff | priceDifferencePercent | qtyDifferencePercent | billingPrice | billingQty | transactionPrice | transactionQty
+    ${"IPV_PASSPORT_CRI_REQUEST_SENT"} | ${"client1"} | ${TimeStamps.CURRENT_TIME} | ${"2"}             | ${"0.0000"} | ${"0"}  | ${"0.0000"}            | ${"0"}               | ${"6.6600"}  | ${"2"}     | ${"6.6600"}      | ${"2"}
+    ${"IPV_PASSPORT_CRI_REQUEST_SENT"} | ${"client1"} | ${TimeStamps.CURRENT_TIME} | ${"1"}             | ${"3.3300"} | ${"1"}  | ${"100.0000"}          | ${"100"}             | ${"6.6600"}  | ${"2"}     | ${"3.3300"}      | ${"1"}
+    ${"IPV_PASSPORT_CRI_REQUEST_SENT"} | ${"client1"} | ${TimeStamps.CURRENT_TIME} | ${undefined}       | ${"6.6600"} | ${"2"}  | ${undefined}           | ${undefined}         | ${"6.6600"}  | ${"2"}     | ${undefined}     | ${undefined}
+  `(
```
markschnitziusgovuk, line 45: There should be an example where billingQty is less than transactionQty, and another where the billingPrice and billingQty are undefined I think.

markschnitziusgovuk (changes requested):

integration_tests/helpers/commonHelpers.ts
NKannan14: fixed

integration_tests/tests/billing-and-transaction-view-tests.ts
NKannan14: added test case of billing qty less than transaction qty

PR #185, opened 2023-01-27
BTM-322 test client id changes

markschnitziusgovuk:
Renames non-prod client ids such as "client1", "client2", ... to match a recent change in config, where these will now be referred to as "vendor_testvendor1", "vendor_testvendor2', ...

This is in anticipation of a later change where we'll be referring to client ids as "vendor ids" from now on.


MelAtGDS (approved):
Assume it was just a find/replace

PR #195, opened 2023-02-02
Btm 357 move event names to config repo

Si3rr4wow:


Si3rr4wow:
Looks like this needs more mocking. One for Friday morning me thinks

src/handlers/int-test-support/helpers/envHelper.ts
```diff
@@ -8,3 +8,12 @@ export const configStackName = (): string => `di-btm-cfg-${configName()}`;
 
 export const runViaLambda = (): boolean =>
   process.env.TEST_VIA_LAMBDA === "true";
+
+export const outputQueue = (): string => {
+  if (
+    process.env.OUTPUT_QUEUE_URL === undefined ||
```
phinterseer, line 14: I'd do this check at the beginning of the handler like in other lambdas.

Also this is a test-helper, so would be good to have another helper in actual production code for this.

src/handlers/int-test-support/helpers/envHelper.ts
```diff
@@ -8,3 +8,12 @@ export const configStackName = (): string => `di-btm-cfg-${configName()}`;
...
+    process.env.OUTPUT_QUEUE_URL.length === 0
```
phinterseer, line 15: And while we're at it, we should refactor our handlers to have an init-phase which is triggered on first function start-up as described [here](https://serverlessfirst.com/function-initialisation/). We could also load the required configs in this init-functions. @gdsgj basically did almost the same thing in his caching approach already and it seems to be a accepted pattern in Lambda to use global code to do some init-stuff. I'd prefer they'd explicitly call an init()-function on startup, but I can live with this approach. :-)

src/shared/utils/config-utils/fetch-event-names.test.ts
```diff
@@ -0,0 +1,21 @@
+import { fetchEventNames } from "./fetch-event-names";
+
+jest.mock("./fetch-vendor-service-config", () => ({
+  fetchVendorServiceConfig: () => [
+    { event_name: "BOUGHT_EGGS" },
+    { event_name: "MADE_CAKE" },
+    { event_name: "ATE_CAKE" },
+    { event_name: "MADE_CAKE" },
+    { event_name: "MADE_TEA" },
+    { event_name: "ATE_CAKE" },
```
phinterseer, line 10: Is there a failure case with 'HAVE_CAKE' and 'EAT_CAKE' somewhere? ;-)

src/shared/utils/config-utils/fetch-vendor-service-config.test.ts
```diff
@@ -0,0 +1,64 @@
...
+    mockedCsvConverterFromString = jest.fn(() => mockedVendorServiceConfig);
+    mockedCsvConverter = { fromString: mockedCsvConverterFromString };
+    mockedGetCsvConverter.mockReturnValue(mockedCsvConverter);
+  });
+
+  test("Throws error if no vendor service config found", async () => {});
+
+  test("Throws error if vendor service config is not valid", async () => {});
+
+  test("Converts CSV to Json", async () => {});
```
phinterseer, line 63: This looks slightly unfinished.

src/handlers/int-test-support/helpers/envHelper.ts
Si3rr4wow: Yeah I was thinking similar things, seems like every handler starts with a block of if this throw that, would be nice to tidy that up.

src/handlers/int-test-support/helpers/envHelper.ts
Si3rr4wow: Oops, I should have checked the directory name 🫢

src/shared/utils/config-utils/fetch-vendor-service-config.test.ts
Si3rr4wow: Filled these in, thanks for catching

src/shared/utils/config-utils/fetch-event-names.test.ts
Si3rr4wow: Hmm I tried it but somehow the test still passes 🤔

src/handlers/int-test-support/helpers/envHelper.ts
Si3rr4wow: I've taken this helper out and moved the check over to the handler. I'll have a think about an init function refactor

MelAtGDS (approved):

PR #196, opened 2023-02-03
BTM-385 refactor csv to event int tests

Si3rr4wow:


integration_tests/tests/transactionCSV-to-s3-event-tests.ts
```diff
@@ -1,97 +1,70 @@
...
-  if (
-    result.Contents === undefined ||
-    result.Contents.length !==
-      testCases.filter((data) => data.expectedPath === "happy").length
-  ) {
-    return false;
-  } else {
-    return true;
-  }
-};
```
Si3rr4wow, line 32: This is the function that used to be passed into waitForTrue. I've taken pretty much the same logic and passed it into this new `poll` function I've made which (I think) has a slightly more straight forward interface

integration_tests/tests/transactionCSV-to-s3-event-tests.ts
```diff
@@ -1,97 +1,70 @@
...
-    // Filter and rename data in the test csv
-    const csvString = objectsToCSV(augmentedData, {
-      filterKeys: ["expectedPath"],
-      renameKeys: new Map([
-        ["givenEntity", "Idp Entity Id"],
-        ["givenLoa", "Minimum Level Of Assurance"],
-        ["givenStatus", "Billable Status"],
-        ["expectedEventId", "Request Id"],
-      ]),
-    });
```
Si3rr4wow, line 52: `addFauxDataToTestCases` (renamed to `augmentTestCases`) and `objectsToCSV` moved into an orchestration function `mockIdpCsvData` to reduce the length of this `beforeAll`

integration_tests/tests/transactionCSV-to-s3-event-tests.ts
```diff
@@ -1,97 +1,70 @@
...
+      data: csv,
+    });
+    await poll<ListObjectsCommandOutput>(
+      async () =>
+        await listS3Objects({
+          bucketName: outputBucket,
+          prefix: transactionsDirectory,
+        }),
+      (result) =>
+        result.Contents?.length !== undefined && result.Contents?.length > 0
```
Si3rr4wow, line 42: Here you can see the two elements of the old `checkS3BucketForEventIds` function. `listS3Objects` is the thing we want to keep doing over and over until we get the desired outcome. We'll know we get our desired outcome when there are items in the `result.Contents` array.

integration_tests/tests/transactionCSV-to-s3-event-tests.ts
```diff
@@ -1,97 +1,70 @@
...
+          const s3Event = JSON.parse(s3Object);
+          expect(s3Event.client_id).toEqual(testCase.expectedClientId);
+          expect(s3Event.event_id).toEqual(testCase.expectedEventId);
+          expect(s3Event.event_name).toEqual(testCase.expectedEventName);
+          break;
+        }
+        case "sad": {
+          expect(s3Object).toEqual("NoSuchKey");
+          break;
+        }
```
Si3rr4wow, line 67: I've bundled both test segments into one loop through the `testCases` array and now we just switch what we're `expect`ing based on the path

src/handlers/int-test-support/helpers/commonHelpers.ts
```diff
@@ -54,6 +54,54 @@ export const waitForTrue = async (
   });
 };
 
+export const poll = async <Resolution>(
+  promise: () => Promise<Resolution>,
+  completionCondition: (resolution: Resolution) => boolean,
```
Si3rr4wow, line 59: Note the generic `<Resolution>` gives you type safety when writing your `completionCondition` function

src/handlers/int-test-support/helpers/mock-data/csv/index.ts
```diff
@@ -1,2 +1,22 @@
-export * from "./addFauxDataToTestCases";
-export * from "./objectsToCsv";
+import { augmentTestCases, TestCases } from "./addFauxDataToTestCases";
+import { objectsToCSV } from "./objectsToCsv";
+import { testCases } from "./transformCSV-to-event-test-data";
+
+export const mockIdpCsvData = (): { testCases: TestCases; csv: Buffer } => {
```
Si3rr4wow, line 5: Like I said above this is just what was happening in the `beforeAll`. I decided to have this function return `{ testCases: TestCases; csv: Buffer }` rather than just `import`ing `testCases` in the test because this made `testCases` feel more tightly coupled to the csv.

NKannan14 (approved):

integration_tests/tests/transactionCSV-to-s3-event-tests.ts
```diff
@@ -1,97 +1,71 @@
...
-        expect(s3Event.client_id).toEqual(testCases[i].expectedClientId);
-        expect(s3Event.event_id).toEqual(testCases[i].expectedEventId);
-        expect(s3Event.event_name).toEqual(testCases[i].expectedEventName);
-      }
-    }
   });
 
-  it("does not store invalid events in the storage/btm-transactions/yyyy-mm-dd folder", async () => {
-    for (let i = 0; i < testCases.length; i++) {
+  it("stores stores the events we care about in the storage bucket", async () => {
```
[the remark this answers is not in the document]
Si3rr4wow: Good catch, thanks Nithya!

PR #203, opened 2023-02-10
Btm 176 grafana dashboard

markschnitziusgovuk:
Fixed view query to return fields we want to show on Grafana, and backs up the JSON version of the Grafana dashboard.


MelAtGDS (approved):
Looks straightforward and the tests are passing which I'll take as a good sign because I definitely need a SQL course to fully understand everything here :)

PR #206, opened 2023-02-13
BTM-363 Derive vendor_id from event_name

MelAtGDS:
✅ Make test event_names unique so they still pass against the config repo changes.
✅ Add function to fetch the vendor_id when given the event_name
✅ Update clean handler to 
        * have knowledge of the incoming payload shape as we can now [self-serve ](https://github.com/alphagov/di-txma-event-self-serve/blob/main/consumer/sqs/billing.json)
        * derive vendor_id if not provided (currently, the transaction csv to json lambda supplies vendor id but incoming TxMA events won't have vendor_id
        * update handler tests 
✅ Update integration tests


src/shared/utils/config-utils/fetch-vendor-id.ts
```diff
@@ -0,0 +1,20 @@
...
+
+export const fetchVendorId = async (eventName: string): Promise<string> => {
+  const configBucket = process.env.CONFIG_BUCKET;
+
+  if (configBucket === undefined)
+    throw new Error("No CONFIG_BUCKET defined in this environment");
+
+  // Event name is unique for each vendor id
+  const vendorServiceConfig = (
+    await fetchVendorServiceConfig(configBucket)
```
markschnitziusgovuk, line 11: I'm a little worried that we're fetching the vendor service config for each and every event, when we should in theory only need to fetch it once.  I think Peter had a recommendation/example where a similar bit of config was lazy-loaded.  Happy to discuss, maybe at standup?

MelAtGDS: @phinterseer Are you able to share the approach you were talking about in dev stand up?

MelAtGDS: Thanks for raising @markschnitziusgovuk ! I've used Gareth's caching approach from `get-vendor-service-config-row` for now. @Si3rr4wow is writing a tech debt ticket to build a wrapper for all functions that will access the vendor config.

src/handlers/clean/handler.ts
```diff
@@ -34,63 +36,39 @@ export const handler = async (event: SQSEvent): Promise<Response> => {
...
 
-  if (
-    process.env.OUTPUT_QUEUE_URL === undefined ||
-    process.env.OUTPUT_QUEUE_URL.length === 0
-  ) {
-    const message = "Output queue URL not set.";
-    console.error(message);
-    throw new Error(message);
+  if (typeof bodyObject !== "object")
+    throw new Error("Event record body not an object.");
```
Si3rr4wow, line 41: do you need this given that `isValidBodyObject` has `typeof x === "object"`?

MelAtGDS: I don't - good spot. Removed now.

markschnitziusgovuk (approved):

PR #212, opened 2023-02-20
More logs to investigate failing tests

NKannan14:
added more console logs to debug the failing test in pipeline. The raw invoice pdf file is uploaded as undefined filename in s3.
 The particular test is passing locally when running against DEV locally whereas failing in pipeline. These logs will be removed once the test is fixed.


MelAtGDS (approved):
Added logs to see why this is failing in the dev pipeline but not when running locally agains the dev stack. This PR will be reverted.

Si3rr4wow (approved):

NKannan14:
Merging this PR eventhough the other test failed due to sdk socket error to investigate raw pdf test

PR #217, opened 2023-02-21
FIX:Removed Receipt.txt in the Billing and transaction curated view

NKannan14:
removed receipt.txt from billing and transaction curated view
Removed standardised lambda test as we have end to end test for invoice 
Increased timeout in invoice end to end test


Si3rr4wow (approved):
good bye file

MelAtGDS:
So timeout has been increased for the `s3-standardised-lambda-test` which now passed but the new failing test is in `s3-invoice-raw-store-test` for the **Unhappy path - Upload invalid pdf to the raw invoice pdf bucket test**.

markschnitziusgovuk (approved):

integration_tests/tests/billing-and-transaction-view-tests.ts
```diff
@@ -9,56 +11,47 @@ import {
...
-  const folderPrefix = "btm_billing_standardised";
-  const testObject: S3Object = {
-    bucket: `${prefix}-storage`,
-    key: `${folderPrefix}/receipt.txt`,
-  };
-  beforeAll(async () => {
+  beforeEach(async () => {
     // tests are enabled to run sequentially as we are deleting the S3 directory in view tests so when running the test
     // in parallel other tests will be interrupted(e.g. sns-s3 tests generate and checks eventId). We can enable to run in parallel
     // once we implement BTM-340 to clean up after each test
```
MelAtGDS, line 36: I can't find BTM-340 in Jira so I don't remember what it is was for and if these comments need to be updated.

integration_tests/tests/billing-and-transaction-view-tests.ts
```diff
@@ -84,22 +77,79 @@ interface TestData {
...
+          s3Object.LastModified !== undefined &&
+          new Date(s3Object.LastModified) >= testStartTime
+      ),
+    {
+      timeout: 50000,
+      nonCompleteErrorMessage:
+        "Invoice data never appeared in standardised folder",
+    }
+  );
+  console.log("**********", res);
```
MelAtGDS, line 146: I think perhaps these three console.logs can be removed now.

integration_tests/tests/billing-and-transaction-view-tests.ts
NKannan14: this is the ticket to improve the cleanup strategy https://govukverify.atlassian.net/browse/BTM-340

integration_tests/tests/billing-and-transaction-view-tests.ts
NKannan14: done

MelAtGDS (approved):

PR #219, opened 2023-02-22
BTM-400 Rename prices to rates

MelAtGDS:
As part of BTM-400 about adding integration tests to the config repo, I would like to change the name of one of the config files from `prices.csv` to `rates.csv`. This necessitates a change in the main app repo as well.

image


MelAtGDS:
The failing test is expected as it is linked to the PR in the config repo which will need to go in first.

phinterseer (approved):

PR #232, opened 2023-03-02
BTM-434 added expiration to buckets

markschnitziusgovuk:
For our Risks and Mitigations compliance, we’ve been asked to ensure data deletion after five years.  This can most easily be accomplished via setting a lifecycle expiration time on the relevant buckets themselves.


Si3rr4wow (approved):
lgtm

MelAtGDS:
Are the `AthenaQueryResultsBucket` and `AthenaInvoiceQueryResultsBucket` temporary stores so therefore don't need to be included in this? 
And, as the `ConfigBucket` only has config files and not any more data, it doesn't need to be set a lifecycle expiration time?

PR #234, opened 2023-03-03
BTM-373 Add new RawInvoice S3 bucket

MelAtGDS:


MelAtGDS:
The only failing test is the `transactionCSV-to-s3-event-tests` which is dependent on BTM-417.

phinterseer (approved):

PR #235, opened 2023-03-03
Don't log everything

Si3rr4wow:


src/handlers/clean/handler.ts
```diff
@@ -57,8 +56,6 @@ async function cleanRecord(record: SQSRecord): Promise<void> {
     vendor_id: vendorId,
   };
 
-  console.log("Cleaned Body Object:", cleanedBodyObject);
```
MelAtGDS, line 60: Can we still log just the vendor_id here? That would still be 'safe' right?

src/shared/utils/sqs.ts
```diff
@@ -9,8 +9,6 @@ export async function sendRecord(
   queueUrl: string,
   messageBody: string
 ): Promise<void> {
-  console.log("sending messageBody " + messageBody);
-
```
MelAtGDS, line 13: Can we log just messageBody.event_id?

NKannan14: one of the int test checks event id in the logs. For debugging purpose can we log eventid ?

src/handlers/store-transactions/handler.ts
```diff
@@ -18,7 +18,6 @@ export const handler = async (event: SQSEvent): Promise<Response> => {
 };
 
 async function storeRecord(record: SQSRecord): Promise<void> {
-  console.log("storing record " + JSON.stringify(record));
```
NKannan14, line 21: one of the int test checks event id in the logs. For debugging purpose can we log eventid ?

src/shared/utils/sqs.ts
Si3rr4wow: yup, done. I've added a condition around it because there's no reason why this should know about events. In theory you could use this function to put anything in a queue. A better solution would be to pass in a list of fields you'd like to log but that seemed like too much work for this PR

src/handlers/store-transactions/handler.ts
Si3rr4wow: I've added it back it but moved it down to just before the event is actually stored. I thought it could be misleading where it was before as we would potentially log that we were storing something but then throw an error before we even tried to store it.

src/handlers/clean/handler.ts
Si3rr4wow: I've gone with event_id as that's what we're logging elsewhere

MelAtGDS (approved):

MelAtGDS:
`transactionCSV-to-s3-event-tests.ts` integration test will pass once BTM-417 is approved and merged in.

PR #246, opened 2023-03-08
Btm 446 regression test suite

NKannan14:
1. Added a new test file called **e2e-tests**  the tests in this file can be run in both lower and higher environments. For higher environments the eventName, vendor id, service_name are derived from config as those values cannot be hard coded. 
2. Reduced number of test cases in **Billing-and-transaction-view test file** as the first 4 test cases are moved into end2end test. The tests in this file cannot be run in higher environment as these tests require specific rate requirement as we have in dev/build config
3. Added function to generate transactionEventsViaFilterLambda
4. Removed all the s3 folder deletion instead delete by events and specific invoice file
5. Enabled parallel run 
6. Changed poll condition to check invoice in standardised bucket based on filename rather than lastModified time


package.json
```diff
@@ -54,6 +54,7 @@
     "husky:install": "husky install",
     "test": "jest --maxWorkers=4",
     "test:integration": "jest --config jest.integration.config.ts --runInBand",
+    "test:e2e": "jest --config jest.e2e.config.ts",
```
phinterseer, line 57: We're not using too many abbreviations normally, so I would not mind calling stuff `endToEnd` instead of `e2e`.

integration_tests/payloads/filterFunctionpayload.json
```diff
@@ -0,0 +1,21 @@
+{
+  "Records": [
+    {
+      "messageId": "a81a7828-579c-4958-b67a-63f81b997f61",
+      "receiptHandle": "AQEBR9preCLGnSoKG64NF+pQ6xkVsST+g3cHYz1MQvP0Xu+0BWv466jrMIVkoMf1eBICI0nBz2eTZaqoxoearK7/3+Nkktgc3Dgo9u7l4r4yRgKHEx6htKHuaOZInpRejZoEjBxWz2sH2B1FONLmo16V6r+8qqCmCeXbfhxKqirpUymb1mTgnTRtQvyAgvexHy+88RQbCxOC2Pcjkg4+qjmaJrW085pyTgtMIdtPAKeFVmxek5PvkgzpYj+vcemBjhJJQQTd1magk2XS4uQvhxe0+vRx6pmFIsCnGjaZHZgWLVDcY6ibA4NA0qxLcgBPt5QAVC5DIg35feTOraAWFD1cHxC3VDfhnAIoSs+HvI6LPljfKK+5C1/c8t2EZYVAyOAIHznQkFpyym0ktEa3/jNHiQ==",
```
phinterseer, line 5: I haven't checked in depth, but does this event need to be so realistic here? We're probably not touching most of these fields in our functions. Maybe put some text in there that indicates that? Like 
`"receiptHandle": "some internal handle"`

package.json
NKannan14: done changed to end2end

integration_tests/payloads/filterFunctionpayload.json
NKannan14: yes no need to be realistic added some text for the unused fields

NKannan14:
I think I can make to one end2end test that can run for both lower level and higher level environments to eliminate the duplication of code. Apologies for making this PR ready to review.

integration_tests/tests/e2e-tests-prodConfig/end2end-tests.ts
```diff
@@ -0,0 +1,240 @@
...
+> => {
+  const configBucket = configStackName();
+
+  const vendorServiceCSV = await getS3Object({
+    bucket: configBucket,
+    key: "vendor_services/vendor-services.csv",
+  });
+  vendorServiceDetails = await csvtojson().fromString(vendorServiceCSV ?? "");
+  return vendorServiceDetails;
+};
```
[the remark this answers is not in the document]
NKannan14: Thanks changed to getVendorServiceConfigRows

integration_tests/tests/e2e-tests-prodConfig/end2end-tests.ts
```diff
@@ -0,0 +1,240 @@
...
+      prefix: "btm_transactions/2022-11-30",
+    });
+    await deleteS3Objects({
+      bucketName,
+      prefix: "btm_transactions/2022-12-01",
+    });
+    await deleteS3Objects({
+      bucketName,
+      prefix: "btm_transactions/2023-01-01",
+    });
```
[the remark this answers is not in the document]
NKannan14: Changed to delete by eventIds and filename and enabled tests to run in parallel

integration_tests/tests/e2e-tests.ts
```diff
@@ -0,0 +1,154 @@
...
+  let invoiceFileName: string;
+  let eventIds: string[];
+  let eventTime: string;
+
+  test.each`
+    testCase                                                                               | eventTime       | transactionQty | billingQty
+    ${"No TransactionQty No TransactionPrice(no events) but has BillingQty Billing Price"} | ${"2022/10/30"} | ${undefined}   | ${"100"}
+    ${"BillingQty BillingPrice equals TransactionQty and TransactionPrice"}                | ${"2022/11/30"} | ${"10"}        | ${"10"}
+    ${"BillingQty BillingPrice greater than TransactionQty and TransactionPrice"}          | ${"2022/12/01"} | ${"10"}        | ${"12"}
+    ${"BillingQty BillingPrice lesser than TransactionQty and TransactionPrice"}           | ${"2023/01/02"} | ${"10"}        | ${"6"}
```
Si3rr4wow, line 38: nice cases 👍

integration_tests/tests/e2e-tests.ts
```diff
@@ -0,0 +1,154 @@
...
+    );
+
+  const billingPrice = unitPrice * billingQty;
+  const transactionPrice = unitPrice * transactionQty;
+  const qtyDifference = billingQty - transactionQty;
+  const priceDifference = billingPrice - transactionPrice;
+  const priceDifferencePercentage = (priceDifference / transactionPrice) * 100;
+  const qtyDifferencePercentage = (qtyDifference / transactionQty) * 100;
+
+  if (billingQty !== undefined && transactionQty !== undefined) {
```
Si3rr4wow, line 120: I think the fact you've had to do this branching here is a clue that maybe these cases don't fit neatly into a `test.each` style

integration_tests/tests/e2e-tests.ts
```diff
@@ -0,0 +1,154 @@
...
+
+    expect(response[0].quantity_difference_percentage).toEqual(
+      String(0 - transactionQty * 100)
+    );
+    expect(response[0].quantity_difference).toEqual(String(-transactionQty));
+    expect(response[0].price_difference).toEqual(
+      String((0 - transactionPrice).toFixed(4))
+    );
+    expect(response[0].price_difference_percentage).toEqual(
+      String(((0 - transactionPrice / transactionPrice) * 100).toFixed(4))
```
Si3rr4wow: Sonar Scan picked up on this, `transactionPrice / transactionPrice` is 1

integration_tests/tests/e2e-tests.ts
```diff
@@ -0,0 +1,154 @@
...
+    expect(response[0].price_difference).toEqual(priceDifference.toFixed(4));
+    expect(response[0].billing_price).toEqual(billingPrice.toFixed(4));
+    expect(response[0].price_difference_percentage).toEqual(
+      priceDifferencePercentage.toFixed(4)
+    );
+  } else if (billingQty !== undefined && transactionQty === undefined) {
+    expect(response[0].billing_quantity).toEqual(billingQty);
+    expect(response[0].billing_price).toEqual(billingPrice.toFixed(4));
+
+    expect(response[0].quantity_difference).toEqual(String(billingQty - 0));
```
Si3rr4wow: why `- 0`?

integration_tests/tests/e2e-tests.ts
```diff
@@ -0,0 +1,154 @@
...
+    expect(response[0].quantity_difference_percentage).toEqual(undefined);
+
+    expect(response[0].price_difference).toEqual((billingPrice - 0).toFixed(4));
```
Si3rr4wow, line 142: why `- 0`?

integration_tests/tests/e2e-tests.ts
```diff
@@ -0,0 +1,154 @@
...
+    expect(response[0].price_difference_percentage).toEqual(undefined);
+  } else if (transactionQty !== undefined && billingQty === undefined) {
+    expect(response[0].transaction_quantity).toEqual(transactionQty);
+    expect(response[0].transaction_price).toEqual(transactionPrice.toFixed(4));
+
+    expect(response[0].quantity_difference_percentage).toEqual(
+      String(0 - transactionQty * 100)
```
Si3rr4wow, line 149: `0 - transactionQty` is the same as just `-transactionQty`

integration_tests/tests/s3-invoice-end-to-end-tests.ts
```diff
@@ -72,19 +65,19 @@ describe("\n Happy path - Upload valid mock invoice pdf and verify data is seen
       ({ Contents }) =>
         !!Contents?.some(
           (s3Object) =>
-            s3Object.LastModified !== undefined &&
-            new Date(s3Object.LastModified) >= testStartTime
+            s3Object.Key !== undefined &&
+            s3Object.Key ===
+              `btm_billing_standardised/${filename.slice(0, 27)}.txt`
```
Si3rr4wow, line 70: where does this 27 come from?

src/handlers/int-test-support/helpers/lambdaHelper.ts
```diff
@@ -32,3 +32,22 @@ export const sendLambdaCommand = async <THelper extends IntTestHelpers>(
     return "Error";
   }
 };
+
+export const invokeLambda = async (
```
Si3rr4wow, line 36: I don't understand this change, can you help me out?

src/handlers/int-test-support/helpers/mock-data/invoice/helpers.ts
```diff
@@ -28,3 +31,57 @@ export const createInvoiceInS3 = async (
     params.filename
   );
 };
+
+export const createInvoiceWithGivenData = async (
```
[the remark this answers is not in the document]
Si3rr4wow: I'll need to get back to you on a suggestion for how to move forward on this because it looks like a few different things hinge on this but I don't like the idea that there's this function that makes invoices and tests the standardisation flow and it's then thrown in all over the place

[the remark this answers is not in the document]
Si3rr4wow: if it works I suppose we should probably go ahead and get this merged so we have end to end tests but we should get something on the backlog to sort this

integration_tests/tests/e2e-tests.ts
```diff
@@ -0,0 +1,154 @@
...
+  serviceName: string
+): Promise<void> => {
+  const tableName = TableNames.BILLING_TRANSACTION_CURATED;
+  const response: BillingTransactionCurated =
+    await queryResponseFilterByVendorServiceNameYearMonth(
+      vendorId,
+      serviceName,
+      tableName,
+      eventTime
+    );
```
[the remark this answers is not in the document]
NKannan14: Thanks, added separate function called calculateExpectedResults

integration_tests/tests/e2e-tests.ts
NKannan14: yes it should not there, removed

integration_tests/tests/e2e-tests.ts
NKannan14: yes removed the 0

src/handlers/int-test-support/helpers/mock-data/invoice/helpers.ts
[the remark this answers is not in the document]
NKannan14: Hi @Si3rr4wow  refactored this function, could you please have a look now?

integration_tests/tests/s3-invoice-end-to-end-tests.ts
Si3rr4wow: ah, how about `.replace('.pdf', '.txt')`

src/handlers/int-test-support/helpers/lambdaHelper.ts
Si3rr4wow: Is the idea to use `sendLambdaCommand` and `invokeLambda` in different circumstances?

NKannan14: sendLamdaCommand is called when the tests run in pipeline I was thinking to reuse this function to trigger the transaction events via filterLambda but the sendLambda has different payload and parameters so added this invokeLambda function which takes only function and payload

Si3rr4wow: >Is the idea to use sendLambdaCommand and invokeLambda in different circumstances? 

⬇️ Below summarises a convo I had with @NKannan14 and some playing around with the code I did to get an understanding ⬇️ 

Yes. `invokeLambda` is used to send an given payload to a given lambda whereas `sendLambdaCommand` sends a given payload to a specific lambda - namely the int-test-support-function. `invokeLambda` facilitates end to end testing by allowing us to kick off the filter lambda directly. We need to invoke the filter directly because [REDACTED] team have the SQS that we subscribe to in their env, so we can't pop things on it ourselves.

integration_tests/tests/e2e-tests.ts
NKannan14: after addingCalculateExpectedResults function the branching looks clean and because of  test.each expect statement wont be repeated

integration_tests/tests/e2e-tests.ts
```diff
@@ -0,0 +1,227 @@
...
+
+const prefix = resourcePrefix();
+const storageBucket = `${prefix}-storage`;
+const standardisedFolderPrefix = "btm_billing_standardised";
+let eventName: string;
+let dataRetrievedFromConfig: TestDataRetrievedFromConfig;
+
+describe("\n Upload invoice to raw invoice bucket and verify billing and transaction_curated view query results matches with expected data \n", () => {
+  beforeAll(async () => {
+    dataRetrievedFromConfig = await retrieveMoreTestDataFromConfig();
```
MelAtGDS, line 33: Why retrieve "more" test data? More than what? Can we be specific about the data being retrieved? Maybe `retrieveVendorServiceAndRatesFromConfig`?

NKannan14: renamed the function

Si3rr4wow (approved):

integration_tests/tests/e2e-tests.ts
NKannan14: @Si3rr4wow Mark is fine with change and agreed to resolve

integration_tests/tests/e2e-tests.ts
NKannan14: @Si3rr4wow Mark is fine with change and agreed to resolve

PR #250, opened 2023-03-13
Untangle business logic

Si3rr4wow:
The goal is to distill the common elements of our lambdas and use them to build up a context which can be passed to business logic. 

Ideally procedures like asserting the presence of env vars or fetching required files from config will become declarative, meaning we can think more about what a lambda does and less about how it does it.

### DIFF
--- a/.eslintrc.yml
+++ b/.eslintrc.yml
@@ -6,6 +6,7 @@ ignorePatterns:
 rules:
   '@typescript-eslint/strict-boolean-expressions': off
   '@typescript-eslint/restrict-template-expressions': off
+  '@typescript-eslint/naming-convention': off
 overrides:
   - files: '*.ts'
     parserOptions:
--- a/src/handler-context/S3ConfigClient.ts
+++ b/src/handler-context/S3ConfigClient.ts
@@ -1,0 +1,28 @@
+import { GetObjectCommand, S3Client } from "@aws-sdk/client-s3";
+
+const getClient = (): S3Client =>
+  new S3Client({
+    region: "eu-west-2",
+    endpoint: process.env.LOCAL_ENDPOINT,
+  });
+
+export class S3ConfigFileClient {
+  private readonly _client: S3Client;
+
+  constructor() {
+    this._client = getClient();
+  }
+
+  public readonly getConfigFile = async (path: string): Promise<string> => {
+    const response = await this._client.send(
+      new GetObjectCommand({
+        Bucket: process.env.CONFIG_BUCKET,
+        Key: path,
+      })
+    );
+    if (response.Body === undefined) {
+      throw new Error(`Config file could not be found at ${path}`);
+    }
+    return await response.Body.transformToString();
+  };
+}
--- a/src/handlers/filter/handler.ts
+++ b/src/handlers/filter/handler.ts
@@ -1,29 +1,44 @@
-import { SQSEvent } from "aws-lambda";
-import { Response } from "../../shared/types";
+import { buildHandler } from "../../handler-context";
+import { ConfigFileNames } from "../../handler-context/Config";
 import { sendRecord } from "../../shared/utils";
-import { fetchEventNames } from "../../shared/utils/config-utils/fetch-event-names";
 
-export const handler = async (event: SQSEvent): Promise<Response> => {
-  const outputQueueUrl = process.env.OUTPUT_QUEUE_URL;
-  if (outputQueueUrl === undefined || outputQueueUrl.length === 0)
-    throw new Error("No OUTPUT_QUEUE_URL defined in this environment");
+enum FilterEnv {
+  OUTPUT_QUEUE_URL = "OUTPUT_QUEUE_URL",
+}
 
-  const response: Response = { batchItemFailures: [] };
+interface FilterableMessage {
+  event_name: string;
+}
 
-  const validEventNames = await fetchEventNames();
+type FilterConfigFiles = ConfigFileNames.services | ConfigFileNames.e2eTest;
 
-  const promises = event.Records.filter((record) => {
-    const body = JSON.parse(record.body);
-    console.log(body.event_name);
-    return Boolean(body?.event_name) && validEventNames.has(body.event_name);
-  }).map(async (record) => {
-    try {
-      await sendRecord(outputQueueUrl, record.body);
-    } catch (e) {
-      response.batchItemFailures.push({ itemIdentifier: record.messageId });
-    }
-  });
+const envVars = [FilterEnv.OUTPUT_QUEUE_URL];
 
-  await Promise.all(promises);
-  return response;
-};
+const messageTypeGuard = (
+  maybeMessage: any
+): maybeMessage is FilterableMessage => !!maybeMessage?.event_name;
+
+const outputs = [
+  { destination: FilterEnv.OUTPUT_QUEUE_URL, store: sendRecord },
+];
+
+const configFiles = [ConfigFileNames.services, ConfigFileNames.e2eTest];
+
+export const handler = buildHandler<
+  FilterableMessage,
+  FilterEnv,
+  FilterConfigFiles
+>({
+  envVars,
+  messageTypeGuard,
+  outputs,
+  configFiles,
+})(async ({ messages, config: { services } }) => {
+  const validEventNames = new Set<string>(
+    (services as Array<{ event_name: string }>).map(
+      ({ event_name }) => event_name
+    )
+  );
+
+  return messages.filter(({ event_name }) => validEventNames.has(event_name));
+});
--- a/package-lock.json
+++ b/package-lock.json
@@ -8,6 +8,7 @@
       "name": "di-billing-transaction-monitoring",
       "version": "0.0.1",
       "devDependencies": {
+        "@aws-lambda-powertools/logger": "^1.6.0",
         "@aws-sdk/client-athena": "^3.188.0",
         "@aws-sdk/client-cloudformation": "^3.188.0",
         "@aws-sdk/client-cloudwatch-logs": "^3.188.0",
@@ -149,6 +150,22 @@
         "@aws-sdk/types": "^3.110.0",
         "@aws-sdk/util-utf8-browser": "^3.0.0",
         "tslib": "^1.11.1"
+      }
+    },
+    "node_modules/@aws-lambda-powertools/commons": {
+      "version": "1.6.0",
+      "resolved": "https://registry.npmjs.org/@aws-lambda-powertools/commons/-/commons-1.6.0.tgz",
+      "integrity": "sha512-83YWf63FZN872Bodk+tJHdSZ2o8y5zsAcpQZgBTqL8TEFNBef/zkDuIuvDIUBmif9q+ZaYVAEtileWjjDTvgGA==",
+      "dev": true
+    },
+    "node_modules/@aws-lambda-powertools/logger": {
+      "version": "1.6.0",
+      "resolved": "https://registry.npmjs.org/@aws-lambda-powertools/logger/-/logger-1.6.0.tgz",
+      "integrity": "sha512-RIx8nHDH8exxPfMhkTnxWZMNOEOgf1CjkzbisLvX89/8Z0tMyruDFHdnolZXQ6Z1AyUE5FVaqN5FWlkmyqxS4A==",
+      "dev": true,
+      "dependencies": {
+        "@aws-lambda-powertools/commons": "^1.6.0",
+        "lodash.merge": "^4.6.2"
       }
     },
     "node_modules/@aws-sdk/abort-controller": {
@@ -15129,6 +15146,22 @@
         "@aws-sdk/types": "^3.110.0",
         "@aws-sdk/util-utf8-browser": "^3.0.0",
         "tslib": "^1.11.1"
+      }
+    },
+    "@aws-lambda-powertools/commons": {
+      "version": "1.6.0",
+      "resolved": "https://registry.npmjs.org/@aws-lambda-powertools/commons/-/commons-1.6.0.tgz",
+      "integrity": "sha512-83YWf63FZN872Bodk+tJHdSZ2o8y5zsAcpQZgBTqL8TEFNBef/zkDuIuvDIUBmif9q+ZaYVAEtileWjjDTvgGA==",
+      "dev": true
+    },
+    "@aws-lambda-powertools/logger": {
+      "version": "1.6.0",
+      "resolved": "https://registry.npmjs.org/@aws-lambda-powertools/logger/-/logger-1.6.0.tgz",
+      "integrity": "sha512-RIx8nHDH8exxPfMhkTnxWZMNOEOgf1CjkzbisLvX89/8Z0tMyruDFHdnolZXQ6Z1AyUE5FVaqN5FWlkmyqxS4A==",
+      "dev": true,
+      "requires": {
+        "@aws-lambda-powertools/commons": "^1.6.0",
+        "lodash.merge": "^4.6.2"
       }
     },
     "@aws-sdk/abort-controller": {
--- a/package.json
+++ b/package.json
@@ -4,6 +4,7 @@
   "version": "0.0.1",
   "type": "module",
   "devDependencies": {
+    "@aws-lambda-powertools/logger": "^1.6.0",
     "@aws-sdk/client-athena": "^3.188.0",
     "@aws-sdk/client-cloudformation": "^3.188.0",
     "@aws-sdk/client-cloudwatch-logs": "^3.188.0",
--- a/src/handler-context/Config.ts
+++ b/src/handler-context/Config.ts
@@ -1,0 +1,137 @@
+import csvToJson from "csvtojson";
+
+export enum ConfigFileNames {
+  rates = "rates",
+  services = "services",
+  renamingMap = "renamingMap",
+  inferences = "inferences",
+  transformations = "transformations",
+  vat = "vat",
+  standardisation = "standardisation",
+  e2eTest = "e2eTest",
+}
+
+export type ConfigFiles = Record<ConfigFileNames, unknown>;
+
+enum FileTypes {
+  csv,
+  json,
+}
+
+const fileMap: Record<ConfigFileNames, { path: string; type: FileTypes }> = {
+  [ConfigFileNames.rates]: {
+    path: "rate_tables/rates.csv",
+    type: FileTypes.csv,
+  },
+  [ConfigFileNames.services]: {
+    path: "vendor_services/vendor-services.csv",
+    type: FileTypes.csv,
+  },
+  [ConfigFileNames.renamingMap]: {
+    path: "csv_transactions/header-row-renaming-map.json",
+    type: FileTypes.json,
+  },
+  [ConfigFileNames.inferences]: {
+    path: "csv_transactions/event-inferences.json",
+    type: FileTypes.json,
+  },
+  [ConfigFileNames.transformations]: {
+    path: "csv_transactions/event-transformation.json",
+    type: FileTypes.json,
+  },
+  [ConfigFileNames.vat]: { path: "uk-vat.json", type: FileTypes.json },
+  [ConfigFileNames.standardisation]: {
+    path: "vendor-invoice-standardisation.json",
+    type: FileTypes.json,
+  },
+  [ConfigFileNames.e2eTest]: { path: "e2e-test.json", type: FileTypes.json },
+};
+
+type Json = string | number | boolean | null | Json[] | { [key: string]: Json };
+
+interface ConfigClient {
+  getConfigFile: (path: string) => Promise<string>;
+}
+
+const parseCsvFile = async (rawFile: string): Promise<Json> => {
+  return await csvToJson().fromString(rawFile);
+};
+const parseJsonFile = (rawFile: string): Json => {
+  return JSON.parse(rawFile);
+};
+
+const parseConfigFile = async (
+  rawFile: string,
+  type: FileTypes
+): Promise<Json> => {
+  switch (type) {
+    case FileTypes.json:
+      return parseJsonFile(rawFile);
+    case FileTypes.csv:
+      return await parseCsvFile(rawFile);
+  }
+};
+
+export class Config<TFileName extends ConfigFileNames> {
+  private readonly _client: ConfigClient;
+  private readonly _files: ConfigFileNames[];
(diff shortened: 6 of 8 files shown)


.eslintrc.yml
```diff
@@ -6,6 +6,7 @@ ignorePatterns:
 rules:
   '@typescript-eslint/strict-boolean-expressions': off
   '@typescript-eslint/restrict-template-expressions': off
+  '@typescript-eslint/naming-convention': off
```
Si3rr4wow, line 9: This guy is always kicking off about `event_name` etc. Bored of it to be honest.

src/handler-context/Config/__mocks__/index.ts
```diff
@@ -0,0 +1,45 @@
+import { ConfigFiles } from "../types";
+
+// TODO: fill in the rest of these
+const config: ConfigFiles = {
```
Si3rr4wow, line 4: this will act as the default mock of the config. If you just do a `jest.mock("src/handler-context/Config/__mocks__/index.ts")` then you'll get this.

src/handler-context/Config/__mocks__/index.ts
```diff
@@ -0,0 +1,45 @@
...
+};
+
+let configOverrides = Object.create(null);
+export const __setConfigOverrides = (overrides: Partial<ConfigFiles>): void => {
+  configOverrides = overrides;
+};
+
+export const __resetOverrides = (): void => {
+  configOverrides = Object.create(null);
+};
```
Si3rr4wow, line 36: this seems a little strange but it's a strategy I took [directly from the docs](https://jestjs.io/docs/manual-mocks#examples)

src/handler-context/Config/index.ts
```diff
@@ -0,0 +1,69 @@
...
+      })();
+      return await promise;
+    });
+  };
+
+  public readonly populateCache = async (): Promise<void> => {
+    const resolutions = await Promise.allSettled(this._promises);
+    // @ts-expect-error
+    this._cache = resolutions.reduce<
+      Omit<ConfigFiles, keyof Omit<ConfigFiles, TFileName>>
```
phinterseer, line 45: This seems like a very complex type. Why is that necessary?

src/handler-context/S3ConfigClient.ts
```diff
@@ -0,0 +1,96 @@
...
+  }
+};
+
+const getClient = (): S3Client =>
+  new S3Client({
+    region: "eu-west-2",
+    endpoint: process.env.LOCAL_ENDPOINT,
+  });
+
+export class S3ConfigFileClient {
```
phinterseer, line 70: Why is this a class?

src/handler-context/Config/index.ts
```diff
@@ -0,0 +1,69 @@
+import { Logger } from "@aws-lambda-powertools/logger";
+import { Json } from "../../shared/types";
+import { ConfigClient, ConfigFileNames, ConfigFiles } from "./types";
+
+export class Config<TFileName extends ConfigFileNames> {
+  private readonly _client: ConfigClient;
```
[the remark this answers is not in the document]
Si3rr4wow: Yeah I suppose there is no reason to be doing this now we have the private keyword. I will remove the `_`.

src/handler-context/Config/index.ts
```diff
@@ -0,0 +1,69 @@
...
+    this._hasCacheBeenPopulated = true;
+  };
+
+  public readonly getCache = (): Omit<
+    ConfigFiles,
+    keyof Omit<ConfigFiles, TFileName>
+  > => {
+    if (!this._hasCacheBeenPopulated) {
+      this._logger.warn(
+        "Called getCache before awaiting populateCache. Consider ensuing the cache is populated before reading it."
```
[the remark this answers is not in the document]
Si3rr4wow: Oh good catch. I'll make the message more assertive

src/handler-context/Config/index.ts
Si3rr4wow: It's kind of like [`Pick`](https://www.typescriptlang.org/docs/handbook/utility-types.html#picktype-keys) but it works with an enum, unfortunately `Pick<ConfigFileNames, keyof TFileName>` is not valid. This type is necessary because it scopes in the type of the cached config to just the requested files, rather than all the files that could possibly be there. So if you build a handler context asking for just the vendor services and the rates configs then the resulting config cache type will be just

```
{
  services: Json;
  rates: Json;
}
```

Si3rr4wow: I did need this in 3 places though, I'll declare it and reuse it

src/handler-context/S3ConfigClient.ts
Si3rr4wow: Fair point, I guess I just had classes on the brain when I was working in this area. I'll make it all functions

src/handler-context/Config/__mocks__/index.ts
Si3rr4wow: yep absolutely right, dud copy and paste from me there 😅

src/handler-context/Config/index.ts
```diff
@@ -0,0 +1,69 @@
+import { Logger } from "@aws-lambda-powertools/logger";
+import { Json } from "../../shared/types";
+import { ConfigClient, ConfigFileNames, ConfigFiles } from "./types";
+
+export class Config<TFileName extends ConfigFileNames> {
+  private readonly _client: ConfigClient;
+  private readonly _files: ConfigFileNames[];
+  private readonly _logger: Logger;
+  private _hasCacheBeenPopulated: boolean = false;
+  private _cache!: Omit<ConfigFiles, keyof Omit<ConfigFiles, TFileName>>;
```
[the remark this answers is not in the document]
Si3rr4wow: yeah, sadly you can't do Pick like this. I guess it isn't able to consider TFileName as a union of strings because it's an extension of an enum or something along those lines.

src/handler-context/Config/index.ts
```diff
@@ -0,0 +1,69 @@
...
+    this._promises = this._files.map(async (fileName) => {
+      const promise = (async (): Promise<{
+        file: Json;
+        fileName: ConfigFileNames;
+      }> => {
+        const file = await this._client.getConfigFile(fileName);
+        return { file, fileName };
+      })();
+      return await promise;
+    });
```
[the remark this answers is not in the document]
Si3rr4wow: lush, stolen, thanks!

src/handler-context/Config/index.ts
```diff
@@ -0,0 +1,69 @@
...
+      Omit<ConfigFiles, keyof Omit<ConfigFiles, TFileName>>
+    >((_config, resolution) => {
+      if (resolution.status === "rejected") throw new Error(resolution.reason);
+      const { file, fileName } = resolution.value;
+      return {
+        ..._config,
+        [fileName]: file,
+      };
+      // @ts-expect-error
+    }, {});
```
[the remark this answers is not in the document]
Si3rr4wow: that is sliiiiick

Si3rr4wow:
Something I forgot to mention: There's 2 approaches to testing the handlers shown here: 

1) For the transaction-csv-to-event-json handler I just test the business logic in isolation. This is really convenient as you don't need to mock any of the services, you just invoke the logic with the context you're expecting.
2) For the filter handler I tested the whole handler. This is less convenient as you need to know a bit about how the context is constructed in order to mock it but it does give you more faith in the way that a given event body will be contextualised. 

Personally I think it's ok for us to test the business logic in isolation, the int tests should catch any issues in parsing particular events into messages. The unit tests on the handler builder should also provide assurances. That said I understand that this is quite a leap.

src/handler-context/Config/index.ts
```diff
@@ -1,69 +1,43 @@
...
-    ConfigFiles,
-    keyof Omit<ConfigFiles, TFileName>
-  > => {
-    if (!this._hasCacheBeenPopulated) {
-      this._logger.warn(
-        "Called getCache before awaiting populateCache. Consider ensuing the cache is populated before reading it."
+  public readonly getCache = (): PickedFiles<TFileName> => {
+    if (this.cache === undefined) {
+      throw new Error(
+        "Called getCache before awaiting populateCache. Ensue the cache is populated before reading it."
```
phinterseer, line 38: ensuRe

src/handler-context/Config/__mocks__/s3-config-client.ts
phinterseer, line 1: More underscore business, this time in directory names. What is this good for again?

src/handler-context/Config/s3-config-client.ts
```diff
@@ -61,25 +61,16 @@ const parseConfigFile = async (
   }
 };
 
-const getClient = (): S3Client =>
-  new S3Client({
-    region: "eu-west-2",
-    endpoint: process.env.LOCAL_ENDPOINT,
-  });
+const client = new S3Client({
+  region: "eu-west-2",
```
phinterseer, line 65: Don't we have the region defined in a central place? If not, it might be a good idea to have it somewhere central and import it from there.

src/handler-context/Config/__mocks__/s3-config-client.ts
Si3rr4wow: Ok this one is not on me, [you can blame Zuckerberg for this](https://jestjs.io/docs/manual-mocks)

src/handler-context/context-builders/index.test.ts
```diff
@@ -0,0 +1,224 @@
...
+              messageId: "msg_2",
+              body: `{"doesn't": "conform"}`,
+            },
+          ],
+        } as SQSEvent,
+        testOptions
+      );
+    } catch (error) {
+      expect((error as Error).message).toContain(
+        "Could not process message msg_2"
```
phinterseer, line 151: This could be a better error message. "Incoming data did not conform to TypeGuard" or similar.

src/handler-context/context-builders/index.test.ts
```diff
@@ -0,0 +1,224 @@
...
+                object: { key: "dud-key" },
+              },
+            },
+          ],
+        } as S3Event,
+        testOptions
+      );
+    } catch (error) {
+      expect((error as Error).message).toContain(
+        "The document this event pertains to could not be found"
```
phinterseer, line 201: "The object this event references could not be found."

src/handler-context/context-builders/index.ts
```diff
@@ -0,0 +1,36 @@
...
+>(
+  event: S3Event | SQSEvent,
+  {
+    envVars,
+    messageTypeGuard,
+    outputs,
+    configFiles,
+  }: CtxBuilderOptions<TMessage, TEnvVars, TConfigFileNames>
+): Promise<HandlerCtx<TMessage, TEnvVars, TConfigFileNames>> => {
+  // for the love of god find a nicer way to do this chain
```
phinterseer: Put a TODO here so that SonarCloud can appropriately harass you. :-)

src/handler-context/Config/index.ts
```diff
@@ -0,0 +1,40 @@
...
+  constructor(client: ConfigClient, files: TFileName[]) {
+    this.client = client;
+    this.files = files;
+    this.promises = this.spawnPromises();
+  }
+
+  private readonly spawnPromises = (): Array<
+    Promise<[ConfigFileNames, Json]>
+  > => {
+    if (process.env.CONFIG_BUCKET === undefined)
```
phinterseer, line 19: This should be something the ConfigClient should care about. This module would ideally not care about where the config comes from.

src/handler-context/context-builders/messages.ts
```diff
@@ -0,0 +1,84 @@
...
+  return messages;
+};
+
+const makeCtxS3Messages = async <TMessage>({
+  Records,
+}: S3Event): Promise<TMessage[]> => {
+  const promises = Records.map(
+    async ({
+      s3: {
+        bucket: { name },
```
phinterseer, line 44: `bucketName` or just `bucket`

src/handler-context/context-builders/messages.ts
```diff
@@ -0,0 +1,84 @@
...
+  messageTypeGuard: (maybeMessage: any) => maybeMessage is TMessage,
+  logger: Logger
+): Promise<TMessage[]> => {
+  switch (discernEventType(event)) {
+    case EventTypes.S3:
+      return await makeCtxS3Messages(event as S3Event);
+    case EventTypes.SQS:
+      return makeCtxSQSMessages(event as SQSEvent, messageTypeGuard, logger);
+  }
+};
```
phinterseer, line 84:
```suggestion
const isS3Event = (event: any): event is S3Event => !!(event as S3Event).Records[0]?.s3;
const isSQSEvent = (event: any): event is SQSEvent => !!(event as SQSEvent).Records[0]?.messageId;

export const makeCtxMessages = async <TMessage>(
  event: S3Event | SQSEvent,
  messageTypeGuard: (maybeMessage: any) => maybeMessage is TMessage,
  logger: Logger
): Promise<TMessage[]> => {
  if (isS3Event(event)) return await makeCtxS3Messages(event);
  if (isSQSEvent(event)) return makeCtxSQSMessages(event, messageTypeGuard, logger);
  throw new Error("Event type could not be determined");
};
```

Tried to avoid the double if using
```suggestion
  switch (true) {
    case isS3Event(event):
      return await makeCtxS3Messages(event);
    case isSQSEvent(event):
      return makeCtxSQSMessages(event, messageTypeGuard, logger);
  }
```

But TypeScript wasn't able to infer the types from that, although it is quite obvious.

src/handler-context/Config/index.ts
phinterseer, line 1: Reading about `index.ts` files and their heritage of being a descendant of `index.html` in the olden days. It seems they are supposed to be used as a collection of exports to be able to easily import stuff from a directory of files. We seem to be using them here as just another file containing code. I personally don't find a collection of exports particularly desirable (why not export where you define?). But using it as a random place for code which is probably implementing something to do with the name of the directory doesn't appeal to me either. Maybe just give them a proper name?

src/handler-context/context-builders/index.ts
```diff
@@ -0,0 +1,36 @@
...
+    addOutputsToCtx(
+      outputs,
+      await addMessagesToCtx(
+        event,
+        messageTypeGuard,
+        addEnvToCtx(envVars, addLoggerToCtx(getBlankCtx()))
+      )
+    )
+  );
+};
```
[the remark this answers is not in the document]
Si3rr4wow: I've reworked this a bit now, hope you like it :)

src/handler-context/context-builders/blank.ts
```diff
@@ -0,0 +1,9 @@
+import { HandlerCtx } from "..";
+import { ConfigFileNames } from "../Config/types";
+
+export const getBlankCtx = <
+  TMessage,
+  TEnvVars extends string,
+  TConfigFileNames extends ConfigFileNames
+>(): HandlerCtx<TMessage, TEnvVars, TConfigFileNames> =>
+  ({} as unknown as HandlerCtx<TMessage, TEnvVars, TConfigFileNames>);
```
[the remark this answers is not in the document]
Si3rr4wow: This file is gone now, no more fake types in the context builder

src/handler-context/Config/s3-config-client.ts
Si3rr4wow: I don't think we do but yes that would be a good thing to enshrine

src/handler-context/Config/types.ts
```diff
@@ -0,0 +1,52 @@
+import { InferenceSpecifications } from "../../handlers/transaction-csv-to-json-event/convert/make-inferences";
+import { Transformations } from "../../handlers/transaction-csv-to-json-event/convert/perform-transformations";
+import { Json } from "../../shared/types";
+
+export enum ConfigFileNames {
```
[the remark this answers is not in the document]
Si3rr4wow: Yeah Gareth made a comment around this object too, I'll reexamine it

src/handler-context/Config/index.ts
Si3rr4wow: That's a fair point, I've not been as neat as I could have been in terms of the file structure. Personally I quite like the export piping thing because it keeps the imports neat but it can be a faff

src/handler-context/context-builders/messages.ts
Si3rr4wow: ah that's much nicer, thanks!

src/handler-context/context-builders/index.ts
Si3rr4wow: Luckily thanks to @gdsgj I removed this comment altogether 😂

src/handler-context/context-builders/index.test.ts
```diff
@@ -0,0 +1,224 @@
...
+              messageId: "msg_2",
+              body: `"invalid": "json"`,
+            },
+          ],
+        } as SQSEvent,
+        testOptions
+      );
+    } catch (error) {
+      expect((error as Error).message).toContain(
+        "Could not process message msg_2"
```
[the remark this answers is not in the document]
Si3rr4wow: Reworked these so they're more descriptive of the two circumstances

src/handler-context/Config/s3-config-client.ts
Si3rr4wow: Added a new ticket to fix this https://govukverify.atlassian.net/browse/BTM-499

src/handler-context/Config/index.ts
Si3rr4wow: agreed, I've moved it

src/handler-context/Config/s3-config-client.ts
```diff
@@ -0,0 +1,87 @@
...
+  endpoint: process.env.LOCAL_ENDPOINT,
+});
+
+export const s3ConfigFileClient: ConfigClient = {
+  getConfigFile: async (fileName: ConfigFileNames): Promise<Json> => {
+    const { path, fileExtension } = fileMap[fileName];
+
+    const response = await client.send(
+      new GetObjectCommand({
+        Bucket: process.env.CONFIG_BUCKET,
```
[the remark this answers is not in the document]
Si3rr4wow: Added to the backlog https://govukverify.atlassian.net/browse/BTM-518

src/handler-context/S3ConfigClient.ts
```diff
@@ -0,0 +1,96 @@
...
+
+enum FileExtensions {
+  csv,
+  json,
+}
+
+const fileMap: Record<
+  ConfigFileNames,
+  { path: string; fileExtension: FileExtensions }
+> = {
```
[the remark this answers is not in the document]
Si3rr4wow: Done :D

src/handler-context/Config/types.ts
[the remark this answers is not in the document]
Si3rr4wow: I've changed the enum to `ConfigElements`  and the what was `ConfigFiles` is now `ConfigCache`. That means we also now have `PickedConfigCache`, the subset of the cache the be loaded, which is the type returned by `Config.getCache`

src/handler-context/Config/index.ts
Si3rr4wow: Ok, done some tidying up, the index files are now exporting only what's required outside of their directory. The handler -context index is now just 
```
export { buildHandler } from "./build-handler";
export { BusinessLogic, HandlerCtx } from "./types";
export { ConfigElements } from "./config/types";
```
These are the only pieces you need to know about from inside the handlers themselves.

src/handlers/filter/handler.ts
```diff
@@ -1,29 +1,12 @@
...
-};
+import { buildHandler, ConfigElements } from "../../handler-context";
+import { sendRecord } from "../../shared/utils";
+import { businessLogic } from "./business-logic";
+import { ConfigCache, Env, Message } from "./types";
+
+export const handler = buildHandler<Message, Env, ConfigCache>({
+  envVars: [Env.OUTPUT_QUEUE_URL],
+  messageTypeGuard: (maybeMessage: any): maybeMessage is Message =>
+    !!maybeMessage?.event_name,
```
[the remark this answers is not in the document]
markschnitziusgovuk: I believe that `!!maybeMessage?.event_name` will already return false if the message is null, but I've added a check that `typeof maybeMessage?.event_name === "string"`.

src/handler-context/context-builders/dynamic/elements/messages.ts
```diff
@@ -0,0 +1,76 @@
+import { Logger } from "@aws-lambda-powertools/logger";
+import { S3Event, SQSEvent } from "aws-lambda";
+import { fetchS3 } from "../../../../shared/utils";
+
+const makeCtxSQSMessages = <TMessage>(
+  { Records }: SQSEvent,
+  messageTypeGuard: (maybeMessage: any) => maybeMessage is TMessage,
```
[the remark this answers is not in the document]
markschnitziusgovuk: fixed

src/handler-context/types.ts
```diff
@@ -0,0 +1,46 @@
...
+  store: UserDefinedOutputFunction<TEnvVars>;
+}>;
+
+export type BusinessLogic<
+  TMessage,
+  TEnvVars extends string,
+  TConfigElements extends ConfigElements
+> = (
+  ctx: HandlerCtx<TMessage, TEnvVars, TConfigElements>
+) => Promise<unknown[]>;
```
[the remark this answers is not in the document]
markschnitziusgovuk: Fixed in next commit.

src/handlers/filter/handler.test.ts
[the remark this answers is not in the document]
markschnitziusgovuk: Tests included in next commit.

src/handler-context/config/config.ts
```diff
@@ -0,0 +1,40 @@
...
+export class Config<TFileName extends ConfigElements> {
+  private readonly getConfigFile: GetConfigFile;
+  private readonly files: ConfigElements[];
+  private readonly promises: Array<Promise<[ConfigElements, Json]>>;
+  private cache: PickedConfigCache<TFileName> | undefined;
+
+  constructor(getConfigFile: GetConfigFile, files: TFileName[]) {
+    this.getConfigFile = getConfigFile;
+    this.files = files;
+    this.promises = this.spawnPromises();
```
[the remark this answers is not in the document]
markschnitziusgovuk: Interesting...  I learned something here.  He calls populateCache right after the constructor, so I expect the promises aren't fulfilled anyway until the await there completes, but it feels to me like the promises shouldn't be created until populateCache is called.

Si3rr4wow:
I can't approve because I opened but in spirit I approve this being merged

PR #251, opened 2023-03-13
BTM-451: fix deployment

gdsgj:
This hopefully fixes deployment, which was failing due to AWS Lambda layers being disallowed by the GDS deployment stack. This just bundles the logger package with the Lambda function's JavaScript file instead of using a layer

### DIFF
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -21,7 +21,7 @@ export default {
     transactionCsvToJsonEvent:
       "./src/handlers/transaction-csv-to-json-event/handler.ts",
   },
-  externals: ["@aws-lambda-powertools/logger", "aws-sdk"],
+  externals: "aws-sdk",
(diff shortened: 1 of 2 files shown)


webpack.config.js
```diff
@@ -21,7 +21,7 @@ export default {
     transactionCsvToJsonEvent:
       "./src/handlers/transaction-csv-to-json-event/handler.ts",
   },
-  externals: ["@aws-lambda-powertools/logger", "aws-sdk"],
+  externals: "aws-sdk",
```
Si3rr4wow, line 24: is this because the sdk is accessible by default in lambdas?

Si3rr4wow: ooooh ok I see what you did here, should have read the PR description

Si3rr4wow (approved):

PR #264, opened 2023-03-27
BTM-473 retry on socket hangup

markschnitziusgovuk:
Some utility functions to facilitate retry and timeout on long-running AWS calls.


src/handlers/int-test-support/helpers/call-wrappers.test.ts
```diff
@@ -0,0 +1,154 @@
...
+  describe("given a call that takes longer than the timeout", () => {
+    it("fails with a timeout error", async () => {
+      try {
+        // only give it a half a second to complete
+        await callWithTimeout(500)(oneSecondCall)();
+      } catch (error) {
+        expect(error).toBeInstanceOf(Error);
+        expect((error as Error).message).toBe("Operation timed out");
+      }
+      expect.hasAssertions();
```
phinterseer, line 33: Why not use expect:
```suggestion
      await expect(callWithTimeout(500)(oneSecondCall)())
        .rejects.toThrowError(new Error("Operation timed out"));
```

markschnitziusgovuk: Implemented that change through the file now with a bit more extra cleanup.

phinterseer (approved):

PR #280, opened 2023-04-17
updated cleanup script

NKannan14:


integration_tests/cleanup-jest-html-reports.sh
```diff
@@ -1,14 +1,11 @@
 #!/bin/bash
-# Set the number of directories to keep
 num_of_dir_to_keep=$1
```
markschnitziusgovuk, line 2: What happens if someone invokes this script without a parameter?  I wouldn't want it to remove everything.  Maybe worth checking that this has a value and exiting if it doesn't.

markschnitziusgovuk (approved):

package.json
```diff
@@ -54,7 +54,7 @@
   "scripts": {
     "husky:install": "husky install",
     "test": "jest --maxWorkers=4",
-    "test:integration": "jest --config jest.integration.config.ts --maxWorkers=4",
+    "test:integration": "jest --config jest.integration.config.ts --maxWorkers=4 && npm run test:cleanup:reports",
```
phinterseer, line 57: This will only run the clean-up if the tests all pass. I don't know if that is intended.

It will also change the return code of the whole `npm run ...` command to always be successful (as the clean-up will always be successful) and might break error detection for failing tests in the pipeline.

To preserve the return value of jest you can do:

`jest --config jest.integration.config.ts --maxWorkers=4 ; RET=$? ; npm run test:cleanup:reports ; exit $RET`

integration_tests/cleanup-jest-html-reports.sh
phinterseer: You could also default to 3 if nothing is given:

`num_of_dir_to_keep=${1:-3}`

phinterseer (approved):

package.json
NKannan14: ah Thanks Peter, updated npm script to preserve the return value and execute cleanup script regardless of test pass/fail

PR #281, opened 2023-04-17
BTM-486 Fix views to start using new properly partitioned folders

MelAtGDS:
- The **TransactionTable** (Name: `btm_transactions_standardised`) and **InvoiceLineItemTable** (Name: `btm_billing_standardised`) now point to the new S3 Storage buckets, `btm_event_data` and `btm_invoice_data` respectively.
- **Year** and **Month** partition keys have been set up.
- SQL in the `btm_transactions_curated` and  `btm_billing_curated` Athena views has been tidied up.
- Integration tests updated to point to the two new S3 Storage buckets.


src/handlers/store-transactions/handler.test.ts
```diff
@@ -14,7 +14,7 @@ beforeEach(() => {
   process.env = { ...OLD_ENV };
   mockPutS3.mockClear();
   process.env.STORAGE_BUCKET = "store";
-  process.env.TRANSACTIONS_FOLDER = "btm_transactions";
+  process.env.LEGACY_TRANSACTIONS_FOLDER = "btm_transactions";
```
Si3rr4wow, line 17: When will we be able to ditch this?

Si3rr4wow (approved):
lgtm, just one question about when the legacy can be cleaned up

src/handlers/store-transactions/handler.test.ts
markschnitziusgovuk: That's the next ticket that's in progress (BTM-487), where all this stuff is removed.

PR #285, opened 2023-04-24
Btm 521 refactor cleanup before tests

NKannan14:
refactored globalSetup function  to delete events generated by tests btm_event_data/2005 folder.  Test events are generated  with year 2005. Before the test run if we delete the old events and invoices it will be easier to debug if there are any test failures at the end of test run.
refactored listS3Objects to use ListObjectsV2Command  supports pagination and can retrieve upto 1000 objects per request.
refactored deleteS3objects and deleteS3Object function to delete events by batch 
refactored tests to generated events with the year 2005


src/handlers/int-test-support/helpers/athenaHelper.ts
```diff
@@ -57,6 +57,8 @@ const getQueryExecutionStatus = async (
   const response = await athenaClient.send(
     new GetQueryExecutionCommand(params)
   );
+
+  console.log(response.QueryExecution?.Status);
```
NKannan14, line 61: I would like to keep this console.log for few days to see the error its logs as I have seen  query failure intermittently when running tests in github actions but not able to replicate locally

src/handlers/int-test-support/helpers/s3Helper.ts
```diff
@@ -132,6 +136,76 @@ const deleteS3Objects = async (
...
+  batchDeleteS3ObjectsByPrefixesInBatchBasic
+);
+
+/* Deletes s3 objects by keys in batches */
+export const batchDeleteS3ObjectsByKeysInBatchBasic = async (
+  bucketName: string,
+  keysToDelete: ObjectIdentifier[],
+  batchSize: number
+): Promise<DeleteObjectsCommandOutput> => {
+  let result: DeleteObjectsCommandOutput = {
```
phinterseer, line 186: Why is this initialised?

src/handlers/int-test-support/helpers/s3Helper.ts
```diff
@@ -132,6 +136,76 @@ const deleteS3Objects = async (
   );
 };
 
+/* Deletes s3 objects by prefixes in batches */
+export const batchDeleteS3ObjectsByPrefixesInBatchBasic = async (
+  bucketName: string,
+  prefixesToDelete: string[],
+  batchSize: number
```
phinterseer, line 143: Why is `batchSize` a parameter? Does any user of this function care what batch size the delete uses? Could we just always use the maximum batch size?

src/handlers/int-test-support/helpers/s3Helper.ts
```diff
@@ -132,6 +136,76 @@ const deleteS3Objects = async (
...
+  for (const prefixToDelete of prefixesToDelete) {
+    let continuationToken: string | undefined;
+
+    do {
+      const listParams = {
+        Bucket: bucketName,
+        Prefix: prefixToDelete,
+        ContinuationToken: continuationToken,
+      };
+      const listResult = await s3Client.send(
```
phinterseer, line 159: Have you thought about the compatibility of this with the int-test-lambda? Both the `listObjects` call and the `deleteObjects` call need to be made compatible to the calling happening through the lambda. Best to extend/generify the functions for list and delete which are already there.

NKannan14: No, thanks added condition runViaLambda in the functions batchDeleteS3ObjectsByKeysInBatch and batchDeleteS3ObjectsByPrefixesInBatch. These functions use DeleteObjectsCommand and ListObjectsV2Command not DeleteObjectCommand/ListObjectsCommand. How can i extend the existing function list and delete?

phinterseer: The old `deleteS3Objcts` already takes a bucket and one prefix to delete objects, but does it not in batches and not for more than 1000, so is actually not really a good implementation in the first place. Your new implementation can overwrite the old one, with the old one then becoming a special case of your new one when you just pass one prefix.

Similarly your new implementation of `ListS3Objects` can replace the old one, as the old one also stops at 1000 objects.

src/handlers/int-test-support/helpers/s3Helper.ts
```diff
@@ -132,6 +146,83 @@ const deleteS3Objects = async (
   );
 };
 
+/* Deletes s3 objects by prefixes in batches */
+export const batchDeleteS3ObjectsByPrefixesInBatchBasic = async (
```
phinterseer: This name is far too long and unnecessarily complicated. No one cares if it uses batches as long as the objects are deleted in the quickest manner.

I propose `deleteS3ObjectsByPrefixes` or even just `deleteS3Objects`.

src/handlers/int-test-support/helpers/s3Helper.ts
NKannan14: refactored as suggested

src/handlers/int-test-support/helpers/s3Helper.ts
```diff
@@ -32,20 +33,44 @@ interface DataAndTarget {
   target: S3Object;
 }
 
+interface deleteS3ObjectByKeys {
```
phinterseer, line 36: `Objects`

same on the interface below.

src/handlers/int-test-support/helpers/s3Helper.ts
```diff
@@ -32,20 +33,44 @@ interface DataAndTarget {
...
+  bucket: string;
+  keysToDelete: ObjectIdentifier[];
```
phinterseer, line 38: I think you can save the `toDelete` as the function name makes that clear.

src/handlers/int-test-support/helpers/s3Helper.ts
```diff
@@ -32,20 +33,44 @@ interface DataAndTarget {
...
+}
+
+interface deleteS3ObjectByPrefixes {
+  bucketName: string;
+  prefixesToDelete: string[];
```
phinterseer, line 43: Same here.

src/handlers/int-test-support/helpers/s3Helper.ts
phinterseer: Also it's not the keys alone, also a potential version inside `ObjectIdentifier`. Why did we change that to `ObjectIdentifier`?

src/handlers/int-test-support/helpers/s3Helper.ts
```diff
@@ -32,20 +33,44 @@ interface DataAndTarget {
   target: S3Object;
 }
 
+interface deleteS3ObjectByKeys {
+  bucket: string;
+  keysToDelete: ObjectIdentifier[];
+}
+
+interface deleteS3ObjectByPrefixes {
+  bucketName: string;
```
phinterseer, line 42: Consistency with other functions would be nice. Just `bucket` should be enough.

src/handlers/int-test-support/helpers/s3Helper.ts
```diff
@@ -32,20 +33,44 @@ interface DataAndTarget {
...
+      Prefix: params.prefix,
+      ContinuationToken: continuationToken,
+    };
+    const listResult = await s3Client.send(
+      new ListObjectsV2Command(listParams)
+    );
+    if (listResult.Contents === undefined) {
+      return objects;
+    }
+    objects.Contents = objects.Contents?.concat(listResult.Contents);
```
phinterseer, line 69: We should not use the AWS datatype where we don't need it. You will never be populating the metadata bit of the type and neither should you. Just return the array of objects.

src/handlers/int-test-support/helpers/s3Helper.ts
```diff
@@ -93,45 +118,82 @@ const putS3ObjectBasic = async (
...
-      object
-    )) as unknown as DeleteObjectCommandOutput;
-
-  const bucketParams = {
-    Bucket: object.bucket,
-    Key: object.key,
+      IntTestHelpers.deleteS3Objects,
+      params
+    )) as unknown as DeleteObjectsCommandOutput;
+  const result: DeleteObjectsCommandOutput = {
```
phinterseer, line 129: I doubt any function will actually use the return value and do something with the errors, but if you want to be thorough and return them, use your own datatype. We should stop bleeding the AWS-types into our own code. It's ports and adapters style to define your own internal interfaces and that's where we wanted to get to.

src/handlers/int-test-support/helpers/s3Helper.ts
```diff
@@ -93,45 +118,82 @@ const putS3ObjectBasic = async (
...
 };
 
-const deleteS3Object = callWithRetryAndTimeout(deleteS3ObjectBasic);
+const deleteS3Objects = callWithRetryAndTimeout(deleteS3ObjectsBasic);
 
-const deleteS3Objects = async (
-  params: BucketAndPrefix
-): Promise<DeleteObjectCommandOutput[]> => {
+/* Deletes s3 objects by keys in batches */
+const deleteS3ObjectBasic = async (
```
phinterseer, line 159: This should be `deleteS3ObjectsBasic` with an `s`.

src/handlers/int-test-support/helpers/s3Helper.ts
```diff
@@ -93,45 +118,82 @@ const putS3ObjectBasic = async (
 
 const putS3Object = callWithRetryAndTimeout(putS3ObjectBasic);
 
-const deleteS3ObjectBasic = async (
-  object: S3Object
-): Promise<DeleteObjectCommandOutput> => {
+const deleteS3ObjectsBasic = async (
```
phinterseer, line 121: This should be `deleteS3ObjectsByPrefix`.

src/handlers/int-test-support/helpers/s3Helper.ts
```diff
@@ -32,20 +33,44 @@ interface DataAndTarget {
   target: S3Object;
 }
 
+interface deleteS3ObjectByKeys {
+  bucket: string;
+  keysToDelete: ObjectIdentifier[];
+}
+
+interface deleteS3ObjectByPrefixes {
```
phinterseer, line 41: Should be `deleteS3ObjectsByPrefix`.

src/handlers/int-test-support/helpers/s3Helper.ts
phinterseer: You can also leave out the `ByKeys` as it's implicit in what an S3 object is. (Also it's not just the keys, but a key and a version)

src/handlers/int-test-support/helpers/s3Helper.ts
```diff
@@ -93,45 +118,82 @@ const putS3ObjectBasic = async (
...
+    $metadata: {},
   };
-  return await s3Client.send(new DeleteObjectCommand(bucketParams));
+  for (const prefixToDelete of params.prefixesToDelete) {
+    const listResult = await listS3Objects({
+      bucketName: params.bucketName,
+      prefix: prefixToDelete,
+    });
+    if (listResult.Contents && listResult.Contents.length > 0) {
+      const keysToDelete = listResult.Contents.map(({ Key }) => ({ Key }));
```
phinterseer, line 140: Is this for cutting out the version information? If yes, why?

src/handlers/int-test-support/helpers/s3Helper.ts
```diff
@@ -93,45 +118,82 @@ const putS3ObjectBasic = async (
...
-  if (result.Contents === undefined) return [];
-
-  return await Promise.all(
-    result.Contents.map(
-      async (item) =>
-        await deleteS3Object({ bucket: params.bucketName, key: item.Key ?? "" })
-    )
-  );
+    )) as unknown as DeleteObjectsCommandOutput;
+  const result: DeleteObjectsCommandOutput = {
```
phinterseer, line 167: Same with the AWS type bleeding into our code here.

src/handlers/int-test-support/helpers/s3Helper.ts
```diff
@@ -93,45 +118,82 @@ const putS3ObjectBasic = async (
...
   };
-  return await s3Client.send(new DeleteObjectCommand(bucketParams));
+  for (const prefixToDelete of params.prefixesToDelete) {
+    const listResult = await listS3Objects({
+      bucketName: params.bucketName,
+      prefix: prefixToDelete,
+    });
+    if (listResult.Contents && listResult.Contents.length > 0) {
+      const keysToDelete = listResult.Contents.map(({ Key }) => ({ Key }));
+      const deleteResult = await deleteS3Object({
```
phinterseer, line 141: This will probably fail now if we have a large number of objects with a single prefix, but let's worry about that when we need it. The "list 1000, then delete 1000" approach was actually better and we could have made it work through the lambda, but I guess that's a bit much for now.

src/handlers/int-test-support/helpers/s3Helper.ts
NKannan14: I used objectIdentifier instead of plain keys it provides both key and version Id of the object where as keys would only delete latest version of the object

src/handlers/int-test-support/helpers/s3Helper.ts
NKannan14: ok will return just array of objects

src/handlers/int-test-support/helpers/s3Helper.ts
NKannan14: done named as DeleteS3Objects

src/handlers/int-test-support/helpers/s3Helper.ts
NKannan14: change the type for keys to string[]

src/handlers/int-test-support/helpers/s3Helper.ts
NKannan14: Agreed to do the type changes in separate PR as the changes will be more need to update all the calling functions which uses listS3Objects and DeleteS3Objects

PR #287, opened 2023-04-24
BTM-436: refactor clean handler

gdsgj:
This rewrites the function that cleans incoming user identity verification events

### DIFF
--- a/src/shared/utils/config-utils/get-vendor-id.test.ts
+++ b/src/shared/utils/config-utils/get-vendor-id.test.ts
@@ -28,7 +28,7 @@ describe("getVendorId", () => {
     expect(result).toEqual("vendor_testvendor1");
   });
 
-  it("throws an error if the event name is not found in the config bucket", () => {
+  it("throws an error if the event name is not found in the given config", () => {
     expect(() =>
       getVendorId("SOME_EVENT", givenVendorServiceConfig)
     ).toThrowError("Event name: SOME_EVENT not found in vendorServiceConfig");


src/handlers/clean/business-logic.ts
```diff
@@ -0,0 +1,39 @@
...
+    event_id,
+    event_name,
+    timestamp,
+    timestamp_formatted,
+    user,
+    vendor_id,
+  },
+  { logger }
+) => {
+  const vendorId = vendor_id ?? (await fetchVendorId(event_name));
```
Si3rr4wow, line 22: Since `fetchVendorId` fetches config could we rewrite it as a sync function that takes the config as an argument and pass in a config from the context here? It looks like this is the only place that `fetchVendorId` is used so shouldn't cause duplication

src/shared/utils/config-utils/get-vendor-id.test.ts
```diff
@@ -0,0 +1,36 @@
...
+      },
+    ];
+  });
+
+  it("returns a vendor_id if the event name is found in the config", () => {
+    const result = getVendorId("VENDOR_1_EVENT_1", givenVendorServiceConfig);
+    expect(result).toEqual("vendor_testvendor1");
+  });
+
+  it("throws an error if the event name is not found in the config bucket", () => {
```
Si3rr4wow, line 31:
```suggestion
  it("throws an error if the event name is not found in the given config", () => {
```

Si3rr4wow (approved):
Clean af 👌

PR #289, opened 2023-04-26
BTM-437 refactor store lambda

MelAtGDS:
Re-writing the function that stores cleaned user identity verification events.

### DIFF
--- a/src/handlers/store-transactions/business-logic.test.ts
+++ b/src/handlers/store-transactions/business-logic.test.ts
@@ -1,5 +1,4 @@
 import { HandlerCtx } from "../../handler-context";
-import { formatDate } from "../../shared/utils";
 import { businessLogic } from "./business-logic";
 import { CleanedEventBody } from "./types";
 
@@ -7,7 +6,9 @@ describe("Store transactions businessLogic", () => {
   let givenCtx: HandlerCtx<any, any>;
   let validIncomingCleanedEventBody: CleanedEventBody;
 
-  beforeEach(() => {
+  it("returns the key to store the transaction data", async () => {
+    // Arrange
+
     givenCtx = {
       config: {},
       logger: {},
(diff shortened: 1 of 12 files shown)


src/handlers/store-transactions/types.ts
```diff
@@ -0,0 +1,20 @@
...
+    transaction_id?: string;
+  };
+}
+
+export enum Env {
+  STORAGE_BUCKET = "STORAGE_BUCKET",
+  EVENT_DATA_FOLDER = "EVENT_DATA_FOLDER",
+}
+
+export type ConfigCache = ConfigElements.services; // ToDo Needed?
```
markschnitziusgovuk, line 20: Address this ToDo?

src/handlers/store-transactions/business-logic.test.ts
```diff
@@ -0,0 +1,36 @@
...
+      timestamp_formatted: "the formatted timestamp",
+      vendor_id: "some vendor id",
+    };
+  });
+
+  it("returns the key to store the transaction data", async () => {
+    // Arrange
+    const expectedDate = new Date(validIncomingCleanedEventBody.timestamp);
+    const yearMonthDayFolder = formatDate(expectedDate, "/");
+    const expectedKey = `${yearMonthDayFolder}/${validIncomingCleanedEventBody.event_id}.json`;
```
markschnitziusgovuk, line 30: This basically duplicates the logic in business-logic.ts, so should always pass...  I'd suggest maybe just hardcoding the expected string.

src/handlers/store-transactions/types.ts
MelAtGDS: Will do :)

src/handlers/store-transactions/business-logic.test.ts
MelAtGDS: Good suggestion. This came from the existing tests from the handler which I copied over. Will hardcode timestamp and the expected string.

PR #290, opened 2023-04-27
Refactor store lambda, change context storage mechanism

Si3rr4wow:


src/handlers/store-transactions/store.test.ts
```diff
@@ -0,0 +1,32 @@
...
+import { putTextS3 } from "../../shared/utils";
+import { store } from "./store";
+import { CleanedEventBody } from "./types";
+
+jest.mock("../../shared/utils/s3", () => ({
+  putTextS3: jest.fn(),
+}));
+
+describe("store", () => {
+  it("stores an message in S3 in a directory based on the event timestamp and a file based on the event id", async () => {
```
phinterseer, line 11: *a message

src/handlers/store-transactions/types.ts
```diff
@@ -0,0 +1,16 @@
+export interface CleanedEventBody {
```
phinterseer, line 1: I could imagine putting the type-guard for this close to the type definition so that you always think of updating both.

phinterseer (approved):

src/handlers/store-transactions/types.ts
Si3rr4wow: yeah that sounds sensible, I've moved the guard into this file

PR #291, opened 2023-05-03
Do not merge: deployable frontend prototype

gdsgj:


webpack.config.js
```diff
@@ -21,18 +22,36 @@ export default {
...
+          from: "./node_modules/govuk-frontend",
+          to: "node_modules/govuk-frontend",
+        },
+        {
+          from: "./src/frontend/views",
+          to: "views",
+        },
+      ],
+    }),
+  ],
```
Si3rr4wow, line 54: can you explain these please?

Si3rr4wow: Ok, I didn’t see a views folder so I guess this doesn’t render anything as is? How come we need to copy the govuk frontend module specifically but not other dependencies?

PR #292, opened 2023-05-03
Btm 536 refactor test helper types

NKannan14:
refactored int test helpers to use own return type rather than aws ones and updated corresponding tests


src/handlers/int-test-support/helpers/athenaHelper.ts
```diff
@@ -116,17 +101,23 @@ const formattedQueryResults = async (
...
-  getQueryExecutionStatus,
-  getQueryResults,
-  waitAndGetQueryResults,
-  formattedQueryResults,
-  queryObject,
+export const queryObject = async (queryId: string): Promise<any> => {
+  const queryResults: StringObject[] = await waitAndGetQueryResults(queryId);
+  const strFromQuery = JSON.stringify(queryResults);
+  return JSON.parse(strFromQuery);
 };
```
Si3rr4wow, line 123: What's this doing?

src/handlers/int-test-support/helpers/athenaHelper.ts
```diff
@@ -116,17 +101,23 @@ const formattedQueryResults = async (
...
+    const result = await getQueryExecutionStatus(queryId);
+    return result?.state?.match("SUCCEEDED") !== null;
+  };
+  await poll(checkState, (state) => state, {
+    timeout: 65000,
+    interval: 5000,
+    notCompleteErrorMessage: "Query did not succeed within the given timeout",
+  });
+  return await getQueryResults(queryId);
 };
```
Si3rr4wow, line 117:
```suggestion
export const waitAndGetQueryResults = async (
  queryId: string
): Promise<StringObject[]> => {
  await poll(
    async () => await getQueryExecutionStatus(queryId),
    (result) => result?.state?.match("SUCCEEDED") !== null,
    {
      timeout: 65000,
      interval: 5000,
      notCompleteErrorMessage: "Query did not succeed within the given timeout",
    }
  );
  return await getQueryResults(queryId);
};
```

Si3rr4wow:
While we're changing types could we stop using `StringObject` and just use `Record<string, string>`?

integration_tests/tests/s3-invoice-end-to-end-tests.ts
```diff
@@ -173,11 +162,18 @@ describe("\n Happy path - Upload valid mock invoice pdf and verify data is seen
...
+      return {
+        vendor_id: item.vendor_id,
+        vendor_name: item.vendor_name,
+        service_name: item.service_name,
+        quantity: item.quantity,
+        price: item.price,
+        year: item.year,
+        month: item.month,
+      };
     });
```
Si3rr4wow, line 176: Why do this map, are there keys on the `item`s that you don't want?

src/handlers/int-test-support/helpers/commonHelpers.ts
```diff
@@ -134,3 +134,31 @@ export const checkS3BucketForEventId = async (
...
+  price_difference_percentage: string;
+}>;
+
+export const mapBillingTransactionCurated = (
+  items: Array<Record<string, string>>
+): BillingTransactionCurated => {
+  return items.map((item) => {
+    return {
+      vendor_id: item.vendor_id,
+      vendor_name: item.vendor_id,
```
Si3rr4wow, line 155: did you mean to put `vendor_id` here?

integration_tests/tests/s3-invoice-end-to-end-tests.ts
NKannan14: types were different so used map, but now changed to generics type as per suggestion

src/handlers/int-test-support/helpers/commonHelpers.ts
NKannan14: typo it should be vendor_name, removed this map used generics as per suggestion

integration_tests/tests/s3-invoice-raw-store-tests.ts
```diff
@@ -35,12 +35,10 @@ describe("\n Unhappy path - Upload invalid pdf to the raw invoice bucket test\n"
...
         prefix: "failed",
       });
-      if (result.Contents === undefined) {
+      if (result === undefined) {
         return false;
       }
-      return result.Contents.some((items) =>
-        items.Key?.includes(rawInvoice.key)
-      );
+      return result.some((items) => items.key?.includes(rawInvoice.key));
```
Si3rr4wow, line 41: I think this might be a hangover from when we had the lint rule that made you do explicit null checks. You could one liner this into 
```
      return !!result && result.some((items) => items.key?.includes(rawInvoice.key));
```

integration_tests/tests/s3-invoice-standardised-line-item-tests.ts
```diff
@@ -258,8 +258,8 @@ export const listS3Keys = async (
   prefix: string
 ): Promise<string[]> => {
   const result = await listS3Objects({ bucketName, prefix });
-  if (result.Contents === undefined) return [];
-  const keys: Array<string | undefined> = result.Contents.map(({ Key }) => Key);
+  if (result === undefined) return [];
+  const keys: Array<string | undefined> = result.map(({ key }) => key);
```
Si3rr4wow, line 262: Looking at `listS3ObjectsBasic`, where `result` ultimately comes from, the return value is initialised as an empty array then each result is pushed into it so there shouldn't be a case where `result` is undefined. 
You can also get away without the explicit typing of `keys`, `const keys = result.map(({ key }) => key);` is fine :)

integration_tests/tests/transactionCSV-to-s3-event-tests.ts
```diff
@@ -34,10 +34,7 @@ describe("Given a csv with event data is uploaded to the transaction csv bucket"
           prefix: `${transactionsDirectory}/2023/01/01/`,
         }),
       (result) => {
-        return (
-          result.Contents?.length !== undefined &&
-          result.Contents?.length === happyPathCount
-        );
+        return result.length !== undefined && result.length === happyPathCount;
       },
```
Si3rr4wow, line 38: you can just do 
```
      (result) => 
        result.length === happyPathCount
      ,
```
since result is always an array

src/handlers/int-test-support/helpers/s3Helper.ts
```diff
@@ -96,82 +108,72 @@ const getS3ObjectBasic = async (
...
     });
-    if (listResult.Contents && listResult.Contents.length > 0) {
-      const keysToDelete = listResult.Contents.map(({ Key }) => Key).filter(
-        (key): key is string => key !== undefined
-      );
-      const deleteResult = await deleteS3Objects({
+    if (listResult && listResult.length > 0) {
+      const keysToDelete = listResult
+        .map(({ key }) => key)
+        .filter((key): key is string => key !== undefined);
```
Si3rr4wow, line 153: this is cool, like this

Si3rr4wow (approved):
Nice! 👍

integration_tests/tests/s3-invoice-raw-store-tests.ts
NKannan14: ah yes, changed to one liner

integration_tests/tests/s3-invoice-standardised-line-item-tests.ts
NKannan14: done, thanks

PR #297, opened 2023-05-15
Fix deployment

gdsgj:
This fixes deployment by preventing code that requires certain permissions from running in the pipeline instead of a Lambda function, because the pipeline does not have the those permissions like the Lambda function does


NKannan14 (approved):
Thanks for fixing the type

PR #301, opened 2023-05-18
increased timeout in checkstandardised fn poll and jest test Timeout

NKannan14:
1. Increased check standardised function timeout further more 
2. Increased jest test timeout in config files
3. removedContents!==undefined filtering in listS3objects in checkStandardised as Contents is expected be an array

### DIFF
--- a/jest.integration.config.ts
+++ b/jest.integration.config.ts
@@ -8,7 +8,7 @@ const config: Config.InitialOptions = {
   testRunner: "jasmine2",
   globalSetup: "./src/handlers/int-test-support/helpers/testSetup.ts",
   verbose: true,
-  testTimeout: 1600000,
+  testTimeout: 220000,
   reporters: [
     "default",
     [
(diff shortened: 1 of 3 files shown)


NKannan14:
Increased timeout massively in checkStandardised but if there is any better way to optimise this please let me know ::)

jest.integration.config.ts
```diff
@@ -8,7 +8,7 @@ const config: Config.InitialOptions = {
   testRunner: "jasmine2",
   globalSetup: "./src/handlers/int-test-support/helpers/testSetup.ts",
   verbose: true,
-  testTimeout: 1600000,
+  testTimeout: 300000,
```
markschnitziusgovuk, line 11: Description of this PR says you're increasing timeout, but here you're decreasing -- just wanted to make sure this value isn't a mistake.

NKannan14: earlier testTimeout was wrong it was supposed to be 160000 now this has been changed to 300000 ms

PR #302, opened 2023-05-19
Fix deployment

gdsgj:
This fixes a failure in a pipeline integration test that blocked deployment. It was caused by the code attempting to pass JS Date objects to the integration test helper Lambda function via JSON, even though Date objects do not serialise to JSON


NKannan14 (approved):
Looks good to me

PR #306, opened 2023-05-25
BTM-584 Fix test failures

NKannan14:
1. Removed the afterEach from standardised-line-item-tests as it was causing query failure when running the tests in parallel and moved the deletion to globalSetup function.
2. Refactored globalSetup function and included deleting raw-invoice and raw-invoice-textract-data bucket objects.
3. Created file called test-constants to  manage the constant values throughout the tests for ease of maintenance.
4. Increased test timeout and checkStandardised poll timeout  to allow more time for the invoice to appear in standardised folder.

### DIFF
--- a/src/handlers/int-test-support/helpers/testSetup.ts
+++ b/src/handlers/int-test-support/helpers/testSetup.ts
@@ -14,11 +14,13 @@ import {
 } from "../test-constants";
 
 export default async function globalSetup(): Promise<void> {
+  console.log("Cleaning up the environment before test execution starts");
   await deleteS3ObjectsAndPoll(STORAGE_BUCKET, `${S3_TRANSACTION_FOLDER}/2005`);
   await deleteS3ObjectsAndPoll(STORAGE_BUCKET, S3_INVOICE_FOLDER);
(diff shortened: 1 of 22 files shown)


src/handlers/int-test-support/helpers/testSetup.ts
```diff
@@ -1,51 +1,76 @@
...
     }
   );
+};
 
-  // Delete objects with prefix "btm_invoice_data"
-  await deleteS3ObjectsByPrefix({
-    bucket: storageBucket,
-    prefixes: ["btm_invoice_data"],
-  });
+const deleteAllObjects = async (bucketName: string): Promise<void> => {
```
Si3rr4wow, line 48: This could fail if the list of objects is too big. [This thread](https://stackoverflow.com/questions/30755129/aws-s3-listobjects-with-pagination) has some good examples of how to paginate through the results

NKannan14: the listS3ObjectsBasic function in s3Helper , handles the pagination using continuation token.

PR #320, opened 2023-06-08
Btm 585 tear down script fix

NKannan14:
created separate function for listingStack resources
replaced DescribeStackResourcesCommand with ListStackResourcesCommand for pagination as in the AWS docs it suggested to ListStackResourcesCommand instead of DescribeStackResourcesCommand for pagination https://awscli.amazonaws.com/v2/documentation/api/latest/reference/cloudformation/describe-stack-resources.html
update type StackResource to StackResourceSummary as StackRescource not exists in ListStackResourcesCommand
Added condition to do tryToDelete

### DIFF
--- a/src/tools/teardown-cf-stack.ts
+++ b/src/tools/teardown-cf-stack.ts
@@ -119,75 +119,81 @@ const deleteBucket = async (
   return await deleteEmptyBucket(resource);
 };
 
+const listAllStackResources = async (
+  stackName: string
+): Promise<StackResourceSummary[]> => {
+  const resources: StackResourceSummary[] = [];
+  let nextToken: string | undefined;
+  do {
+    await wait(10000);
+    const result = await cfClient
+      .send(
+        new ListStackResourcesCommand({
+          StackName: stackName,
+          NextToken: nextToken,
+        })
+      )
+      .catch((err) => {
+        return { error: err };
+      });
+    if (isAWSError(result)) {
+      if (
+        (result.error as string)
+          .toString()
+          .includes(`Stack with id ${stackName} does not exist`)
+      ) {
+        console.log(`Stack ${stackName} successfully destroyed.`);
+        process.exit(0);
+      } else {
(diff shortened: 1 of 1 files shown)


markschnitziusgovuk (approved):
LGTM

src/tools/teardown-cf-stack.ts
```diff
@@ -149,20 +149,44 @@ const destroyStack = async (stackName: string): Promise<StackResource[]> => {
...
+): Promise<StackResourceSummary[]> => {
+  const deleteResult = await cfClient.send(
+    new DeleteStackCommand({ StackName: stackName })
+  );
+  if (isAWSError(deleteResult)) {
+    console.log(deleteResult.error.toString());
+    process.exit(-1);
+  }
+  let resourcesFailed: StackResourceSummary[] = [];
+  while (true) {
```
Si3rr4wow, line 174: For anyone else who gets spooked seeing this here (or even to myself if I circle back to this) this look is broken by invoking `process.exit`

src/tools/teardown-cf-stack.ts
```diff
@@ -149,20 +149,44 @@ const destroyStack = async (stackName: string): Promise<StackResource[]> => {
...
+  stackName: string
+): Promise<StackResourceSummary[]> => {
+  const deleteResult = await cfClient.send(
+    new DeleteStackCommand({ StackName: stackName })
+  );
+  if (isAWSError(deleteResult)) {
+    console.log(deleteResult.error.toString());
+    process.exit(-1);
+  }
+  let resourcesFailed: StackResourceSummary[] = [];
```
Si3rr4wow, line 173: I think you can get rid of this declaration and keep the old `const resourcesFailed = result.StackResources.filter(` one. It doesn't seem like you need access to this `resourcesFailed` outside the `while` loop

NKannan14: Yes, thanks Mark. Changed to local variable

PR #322, opened 2023-06-09
BTM-575: email decryption etc

gdsgj:
This
- adds automatic decryption for fetched S3 objects that were encrypted using a KMS key and the default SES encryption method
- removes the temporary email bucket
- replaces the temporary bucket's connections to other resources with connections to the actual email bucket
- fixes some minor issues:
  - I ran `npm audit fix` after running NPM showed a message about several "severe" issues
  - I fixed a strange problem where passing an empty string as the S3 endpoint to the S3 client started causing errors, though it was apparently working previously. Maybe `npm audit fix` updated the AWS SDK version in use there, and maybe the upgrade included a breaking change, but I do not know
  - I added a package dependency for an import that started to fail after `npm audit fix`. My guess is that it removed the package as a dependency of another package that was never directly listed as a dependency of this project

### DIFF
--- a/src/shared/utils/kms/index.ts
+++ b/src/shared/utils/kms/index.ts
@@ -1,10 +1,11 @@
+import { KMS } from "aws-sdk";
 import { Blob } from "node:buffer";
 import { TextEncoder } from "node:util";
 import { kms } from "./client";
 
 export const decryptKms = async (
   encryptedBytes: Uint8Array,
-  context: {}
+  context: KMS.EncryptionContextType
 ): Promise<Uint8Array> => {
   const request = {
     CiphertextBlob: encryptedBytes,


cloudformation/cleaning.yaml
```diff
@@ -50,7 +50,8 @@ Resources:
       Environment:
         Variables:
           OUTPUT_QUEUE_URL: !Ref StorageQueue
-          LOCAL_ENDPOINT: !If [IsLocal, 'http://localhost:4566', '']
+          LOCAL_ENDPOINT:
+            !If [IsLocal, 'http://localhost:4566', !Ref AWS::NoValue]
```
Si3rr4wow, line 54: 🧠

package.json
```diff
@@ -74,6 +74,7 @@
   },
   "dependencies": {
     "@aws-lambda-powertools/logger": "^1.6.0",
+    "@aws-sdk/util-utf8-node": "^3.259.0",
```
Si3rr4wow, line 77: wow, I think that's the highest minor version number I've ever seen :o

src/shared/utils/kms/index.ts
```diff
@@ -0,0 +1,29 @@
+import { Blob } from "node:buffer";
+import { TextEncoder } from "node:util";
+import { kms } from "./client";
+
+export const decryptKms = async (
+  encryptedBytes: Uint8Array,
+  context: {}
```
Si3rr4wow, line 7: Be cautious with using `{}` as a type, it's quite _flexible_

Screenshot 2023-06-12 at 11 27 20

src/shared/utils/s3.ts
```diff
@@ -52,12 +62,80 @@ export async function fetchS3(bucket: string, key: string): Promise<string> {
...
+
+  const encryptedKeyBuffer = Buffer.from(encryptedKey, "base64");
+
+  const keyContext = JSON.parse(keyContextText);
+  if (typeof keyContext !== "object" || keyContext === null)
+    throw Error("Invalid key context");
+
+  const key = await decryptKms(encryptedKeyBuffer, keyContext);
+
+  const authTagLength = parseInt(authTagBitCountText, 10) / 8;
```
markschnitziusgovuk, line 103: More a question than a comment...  This looks like it can easily take on a non-integer value, but you then use it in slice().  From what I've googled, slice will internally coerce it to an integer, but are you confident this will always work as expected?

Si3rr4wow (approved):
Super clean integration, v well done! 👏

PR #325, opened 2023-06-13
Fix email PDF attachments

gdsgj:
This fixes two bugs:
- Some invoice emails' PDFs stored in S3 could not be read. This was somehow caused by converting PDF binary data to a string (with UTF-8 encoding, by default) and then back again
- Storing PDF Textract data sometimes timed out or prematurely stopped due to memory usage, so I just increased the allocated memory

### DIFF
--- a/cloudformation/raw-invoice-textract-data-store.yaml
+++ b/cloudformation/raw-invoice-textract-data-store.yaml
@@ -145,6 +145,7 @@ Resources:
             Queue: !GetAtt RawInvoiceTextractDataQueue.Arn
       Handler: storeRawInvoiceTextractData.handler
       KmsKeyArn: !GetAtt KmsKey.Arn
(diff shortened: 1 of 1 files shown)


markschnitziusgovuk (approved):

Si3rr4wow (approved):
yeah that'll do it, nice one!

cloudformation/raw-invoice-textract-data-store.yaml
```diff
@@ -145,6 +145,7 @@ Resources:
             Queue: !GetAtt RawInvoiceTextractDataQueue.Arn
       Handler: storeRawInvoiceTextractData.handler
       KmsKeyArn: !GetAtt KmsKey.Arn
+      MemorySize: 256
```
Si3rr4wow, line 148: How come this?

PR #344, opened 2023-06-23
BTM-599: test email

gdsgj:
This tests that email invoice attachments are stored in the raw invoice bucket


MelAtGDS (approved):
Looks good. I guess the ACs in the ticket need changing - 4 & 5 still refer to non pdfs and csvs.

PR #360, opened 2023-07-05
BTM-646 Refactor website routes - part 1

markschnitziusgovuk:


src/frontend/pages.ts
```diff
@@ -0,0 +1,51 @@
...
+
+export const getPagePath = (page: Page): string => {
+  return page.parent ? getPathRecursive(page) : "/";
+};
+
+const getPathRecursive = (page: Page): string => {
+  return (
+    (page.parent ? getPathRecursive(page.parent) + "/" : "") + page.relativePath
+  );
+};
```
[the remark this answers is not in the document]
markschnitziusgovuk: You'd think, but then (say) the contract page, next level down, generates as "//contracts".

I tried and failed to make it a single method -- the easiest way to see the problem is to look at the number of /'s you need to generate: 1 for home, 1 for its children, 2 for the grandchildren, 3 for great-grandchildren...  So any method that always adds a / is bound to fail, if you want it to work for the home page as well.  That's why I effectively treat the home page as a special case.

I went around on this a few times -- if you can figure out a way to do it in a single method, I'd like to see it...

src/frontend/handlers/contracts.ts
```diff
@@ -1,8 +1,8 @@
-import { RequestHandler } from "express";
 import { getContracts } from "../config";
+import { PageParamsGetter } from "../pages";
 
-export const contractsHandler: RequestHandler = async (_, response) => {
-  response.render("contracts.njk", {
+export const contractsParamsGetter: PageParamsGetter<any> = async (_) => {
```
Si3rr4wow, line 4: why the explicit `any`?

src/frontend/handlers/invoices.ts
```diff
@@ -1,24 +1,22 @@
...
-export const getInvoicesHandler: RequestHandler<
-  unknown,
-  unknown,
-  unknown,
-  { contract_id: string }
-> = async (request, response) => {
+export const invoicesParamsGetter: PageParamsGetter<{
+  contract_id: string;
+}> = async (request) => {
+  console.log(request);
```
Si3rr4wow, line 8: temporary?

src/frontend/pages.ts
```diff
@@ -0,0 +1,78 @@
+import { Request, RequestHandler } from "express";
+import { contractsParamsGetter } from "./handlers/contracts";
+import { invoicesParamsGetter } from "./handlers/invoices";
+import { invoiceParamsGetter } from "./handlers/invoice";
+import path from "node:path";
+
+export type PageParamsGetter<TParams> = (
+  request: Request<TParams, unknown, unknown, unknown>
+) => object;
```
Si3rr4wow, line 9: Maybe unnecessary but it would be nice to have a generic for the return type too, rather than just `object`, which will let you do weird stuff like 

```
type Thing = {
  something: object
}

const thing: Thing = {
  something: Number
}
```

src/frontend/handlers/contracts.ts
markschnitziusgovuk: It complains without it.  Maybe I should specify it explicitly?

markschnitziusgovuk: Oh, right, it should just be {} here I think.

src/frontend/pages.ts
markschnitziusgovuk: I've added return types now, which seem to tidy things up pretty well.

MelAtGDS (approved):

PR #362, opened 2023-07-07
Refactor invoice - check with Mark S

MelAtGDS:


src/frontend/handlers/invoice.ts
```diff
@@ -37,59 +21,22 @@ export const getInvoiceHandler: RequestHandler<
...
       vendorName: config.vendorName,
       contractName: config.contractName,
       contractId: request.query.contract_id,
       year: request.query.year,
       prettyMonth: MONTHS[Number(request.query.month) - 1],
       month: request.query.month,
     },
+    bannerClass: invoiceBanner.bannerClass,
+    invoiceStatus: invoiceBanner.status,
     reconciliationRows,
```
markschnitziusgovuk, line 40: Maybe the `invoice: {...}` wrapper can go away?  I put it in to try to keep things tidy, but the page is for showing an invoice so all these variables are going to pertain to the invoice.  Or everything should go inside it.  I don't feel strongly either way.

src/frontend/utils/status-label.ts
```diff
@@ -0,0 +1,23 @@
+export class StatusLabel {
```
Si3rr4wow, line 1: I know you've just moved this to here, just commenting because I missed when this went in: it seems extreme to use a class for this rather than encoding these as 
```
const statusLabels: {
  STATUS_LABEL_WITHIN_THRESHOLD: {
    message: "Within Threshold",
    class: "govuk-tag--green"
  },
  ...
}
```

MelAtGDS: For consistency, this was the same as the `percentage-discrepancy-special-case` used in the invoice banner. I'll change both.

PR #364, opened 2023-07-11
BTM-648 breadcrumbs

markschnitziusgovuk:
Standardises breadcrumbs across all pages that have them (i.e. everything but the home page).

They are passed in to the Nunjucks pages via a 'breadcrumbData' parameter.  This data is built up using the page definitions added in BTM-646, which may rely on various parameters required to build the page title and path for the breadcrumb link.


src/frontend/handlers/invoice.ts
```diff
@@ -29,6 +29,13 @@ export const invoiceParamsGetter: PageParamsGetter<
   const reconciliationRows = getReconciliationRows(lineItems);
 
   return {
+    pageTitle:
+      config.vendorName +
+      " " +
+      MONTHS[Number(request.params.month) - 1] +
+      " " +
+      request.params.year +
+      " Invoice",
```
Si3rr4wow, line 38: maybe `${config.vendorName} ${MONTHS[Number(request.params.month) - 1]} ${request.params.year} Invoice`?

Si3rr4wow (approved):
Really cool!

src/frontend/handlers/invoice.ts
markschnitziusgovuk: Fixed.

PR #366, opened 2023-07-13
BTM-663 UI homepage navigation panel

MelAtGDS:
UI for the homepage navigation panel as designed in the [moqup here](https://app.moqups.com/DpSOLhFxFMD222IpkWjpPj8BHy6BhoyE/edit/page/ac4619028).

James has 👍🏼 already.


Si3rr4wow (approved):
Nothing to complain about there. Weird class names though, I've never seen them with an ! in before

PR #367, opened 2023-07-13
Fix email test

gdsgj:
This fixes an email test that blocked deployment

The issue had two causes:
1. the test helper did not have permission to decrypt the email decryption key
2. SES changes email message IDs when sent via SMTP instead of the SES API, but the test assumed they would stay the same


markschnitziusgovuk (approved):
lgtm

PR #375, opened 2023-07-18
remove SAM CLI version specification

NKannan14:


Si3rr4wow (commented):
> [This is the recommended approach on supported platforms. It does not require Python to be installed, and is faster than the default installation method.](https://github.com/aws-actions/setup-sam#use-installer)

Nice 👍

Si3rr4wow (approved):

PR #376, opened 2023-07-19
BTM-702: authorisation-failed page

gdsgj:
This adds a page for users to see when they are not on our allowlist


MelAtGDS:
Looks good!
image

Also, this isn't from any change on your PR but do you know why the currency symbol has changed???
image

MelAtGDS (approved):

PR #381, opened 2023-07-20
BTM-647: refactor config getting

gdsgj:
Minimises fetches in site loads without needing devs to get config before it is needed

Included:
- move newer config fetching code to shared folder
- add `getConfig`: uses newer config fetching and caches results
- replace older config fetching and caching


src/shared/utils/config-utils/get-config.ts
```diff
@@ -0,0 +1,19 @@
...
+
+export const getConfig = async <T extends ConfigElements>(
+  fileName: T
+): Promise<ConfigCache[T]> => {
+  const cachedPromise = cachedConfigPromisesByFile[fileName];
+  if (cachedPromise !== undefined) return await cachedPromise;
+  const promise = getConfigFile(fileName);
+  (cachedConfigPromisesByFile[fileName] as Promise<ConfigCache[T]>) = promise;
+  return await promise;
+};
```
Si3rr4wow, line 19: Found a way to avoid coercion :D cache the files not the promises

```suggestion
export type CachedConfigByElement = Partial<{
  [T in ConfigElements]: ConfigCache[T];
}>;

const cachedConfigByFile: CachedConfigByElement = {};

export const getConfig = async <T extends ConfigElements>(
  fileName: T
): Promise<ConfigCache[T]> => {
  const cachedFile = cachedConfigByFile[fileName];
  if (cachedFile !== undefined) return cachedFile;
  const configFile = await getConfigFile(fileName);
  cachedConfigByFile[fileName] = configFile;
  return configFile;
};
```

Si3rr4wow: ok not this, this'll make a new promise if one is already in flight

Si3rr4wow (approved):
🪓

PR #386, opened 2023-07-21
BTM-649: refactor webpage links

gdsgj:
Lets us update page titles and URLs and link HTML in one place and update links automatically

### DIFF
--- a/src/frontend/pages.ts
+++ b/src/frontend/pages.ts
@@ -11,12 +11,7 @@ import { invoicesParamsGetter, invoicesTitleGetter } from "./handlers/invoices";
 import { invoiceParamsGetter, invoiceTitleGetter } from "./handlers/invoice";
 import { homeParamsGetter, homeTitleGetter } from "./handlers/home";
 import path from "node:path";
-import {
-  Contract,
-  Period,
-  ReconciliationRow,
-  OverviewRow,
-} from "./extract-helpers";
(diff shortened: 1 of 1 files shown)


src/frontend/pages.ts
```diff
@@ -39,14 +40,14 @@ export const getRoute = <TParams, TReturn>(
...
+  requestParams: TParams
 ): string => {
   const parentRoute = page.parent ? getRoute(page.parent) : "/";
   let url = path.join(parentRoute, page.relativePath);
   const regex = /:([A-Za-z_]+)/;
-  const params = request.params as Record<string, string>;
+  const params = requestParams as Record<string, string>;
   while (regex.exec(url)) {
     url = url.replace(regex, (a, b) => params[b]);
   }
```
markschnitziusgovuk, line 53: I do worry here about params[b] not existing -- we assume it's there but there's no checking.  This is an existing issue (that I introduced, but couldn't think of a better way) -- maybe we should throw an error at least instead of formatting an 'undefined' into the url silently?

src/frontend/handlers/home.ts
```diff
@@ -1,13 +1,22 @@
-import { IndexParams, PageParamsGetter, PageTitleGetter } from "../pages";
+import {
+  IndexParams,
```
markschnitziusgovuk, line 2: Consider renaming everything from 'index' to 'home' for consistency.

markschnitziusgovuk (approved):

PR #387, opened 2023-07-24
BTM-638: error page

gdsgj:
Shows page on `500` errors

Also adds environment variable `POWERTOOLS_SERVICE_NAME` to show Amazon Lambda function name in error logs

### DIFF
--- a/src/frontend/pages.test.ts
+++ b/src/frontend/pages.test.ts
@@ -2,35 +2,45 @@ import { NextFunction, Request, RequestHandler, Response } from "express";
 import { getHandler, getRoute, Page } from "./pages";
 
 describe("Page", () => {
-  const homePage: Page<{}, {}> = {
-    relativePath: "",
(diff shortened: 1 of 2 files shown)


src/frontend/views/error.njk
```diff
@@ -0,0 +1,13 @@
+{% extends "common/layout/base.njk" %}
+{% set mainClasses = "govuk-main-wrapper--l" %}
+{% block pageTitle %}
+  Sorry, there is a problem with the service – Billings and Reconciliation – GOV.UK
```
markschnitziusgovuk, line 4: Should maybe be a separate ticket, but we should probably have a pageTitleFormatter() function or something that takes the main text you want to put in the title and adds a standard ending to it (like " -- Billings and Reconciliation -- GOV.UK").

markschnitziusgovuk (approved):

PR #391, opened 2023-07-25
added link for cookies to footer

Xfarya:
Added cookies link to base.njk - cookies link should appear in footer

### DIFF
--- a/src/frontend/views/common/layout/base.njk
+++ b/src/frontend/views/common/layout/base.njk
@@ -1,4 +1,5 @@
 {% extends "govuk/template.njk" %}
+{% from "govuk/components/footer/macro.njk" import govukFooter %}
 {% block head %}
   <link href="/assets/styles/all.css" rel="stylesheet">
 {% endblock head %}
@@ -9,6 +10,7 @@
 {% endblock pageTitle %}
 {% block content %}
 {% endblock content %}
+{% block footer %}{{ govukFooter({"meta": {"items": [{"href": "#", "text": "Cookies"}]}}) }}{% endblock %}
 {% block bodyEnd %}
   <script src="/assets/scripts/all.js" nonce="{{ cspNonce }}"></script>
   <script>window.GOVUKFrontend.initAll();</script>
--- a/src/frontend/handlers/__snapshots__/authorisation-failed.test.ts.snap
+++ b/src/frontend/handlers/__snapshots__/authorisation-failed.test.ts.snap
@@ -100,11 +100,25 @@ exports[`authorisation failed handler Page displays title and info 1`] = `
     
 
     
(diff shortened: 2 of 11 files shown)


src/frontend/views/cookies.njk
```diff
@@ -0,0 +1,37 @@
...
+        <thead class="govuk-table__head">
+          <tr class="govuk-table__row">
+            <th class="govuk-table__header" scope="col">Name</th>
+            <th class="govuk-table__header" scope="col">Purpose</th>
+            <th class="govuk-table__header" scope="col">Expires</th>
+          </tr>
+        </thead>
+        <tbody class="govuk-table__body">
+          <tr class="govuk-table__row">
+            <td class="govuk-table__cell">auth_token</td>
```
markschnitziusgovuk, line 29: MarkP -- is this an okay name for the actual cookie?

src/frontend/views/error.njk
```diff
@@ -1,8 +1,5 @@
 {% extends "common/layout/base.njk" %}
 {% set mainClasses = "govuk-main-wrapper--l" %}
-{% block pageTitle %}
-  {{ headTitle }}
-{% endblock pageTitle %}
```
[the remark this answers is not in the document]
markschnitziusgovuk: I noticed the error.njk inherits from base.njk, so it was generating the page title twice, which is why I removed it here.  And as that would mean that base.njk is the only place that generates the title now, I thought it made more sense to format the page title there instead of in that function (which I meant to delete -- will do that now).

src/frontend/views/common/layout/base.njk
```diff
@@ -9,6 +10,11 @@
 {% endblock pageTitle %}
 {% block content %}
 {% endblock content %}
+{% block footer %}
+  {# djlint:off #}
+  {{ govukFooter({ meta: { items: [ cookiesLink ] } } ) }}
+  {# djlint:on #}
```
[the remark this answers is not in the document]
markschnitziusgovuk: Fixed.  For some reason whenever I couldn't figure out how to make both lint and govukFooter happy, but this seems to work.

src/frontend/views/cookies.njk
markschnitziusgovuk: Gareth later commented: "I thought there were two different cookies: Access-Token and Refresh-Token. Feel free to mark this as resolved if requirements changed without a ticket update".  Do I use those names here?  If so, what should the short description for each read?

Si3rr4wow: yep, just the one cookie needed with the way the auth is working out

Si3rr4wow (approved):

PR #394, opened 2023-07-25
BTM-518: environment variable helper

gdsgj:
Adds a function to get environment variables and uses it


phinterseer (approved):
When I did a similar thing on Cazoo, I made the variable name parameter of the `getFromEnv` function strongly typed, so that only a fixed set of strings could be used as keys. This protects from accessing random env-vars and gives you nice code-completion in those string literals.

phinterseer:
Yes, pretty much that.

PR #395, opened 2023-07-25
BTM-324: UK dates

gdsgj:
Puts dates in UK time zone instead of UTC, to match transaction data

### DIFF
--- a/integration_tests/tests/athena-query-tests.ts
+++ b/integration_tests/tests/athena-query-tests.ts
@@ -1,8 +1,18 @@
 import {
   invalidEventPayloadEventName,
   validEventPayload,
-  validEventPayloadOneHourBeforeJanuaryUtc,
-  validEventPayloadOneHourBeforeAugustUtc,
+  validUtc6EventPayload1h1sIntoOctUtc1,
+  validUtc6EventPayload1sIntoFebUtc0,
+  validUtc6EventPayload1sIntoOctUtc1,
+  validUtc6EventPayload59m59sIntoOctUtc1,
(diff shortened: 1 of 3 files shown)


Si3rr4wow (commented):
Could we add some integration level testing to this too? Just a couple of events with relevant timestamps, could probably just tweak the existing test data

integration_tests/tests/athena-query-tests.ts
```diff
@@ -16,6 +18,28 @@ describe("\nGenerate valid event and execute athena query\n", () => {
...
+describe("\nGenerate valid event one hour before January UTC and execute athena query\n", () => {
+  test("should not contain eventId in the generated query results for January", async () => {
+    const { eventId } = await invokeFilterLambdaAndVerifyEventInS3Bucket(
+      validEventPayloadOneHourBeforeJanuaryUtc
+    );
+    const queryString = `SELECT * FROM "btm_transactions_standardised" where month='01'`;
+    const queryResult = await queryAthena(queryString);
+    expect(JSON.stringify(queryResult)).not.toContain(eventId);
+  });
+});
```
markschnitziusgovuk, line 41: Considering also adding an event that's just outside the target month, and testing that it doesn't get returned.

markschnitziusgovuk: Sorry, that comment was meant for the preceding test.  But both tests maybe could maybe stand to have an 'inside' and 'outside' case.

markschnitziusgovuk: Resolved in dev standup.

PR #399, opened 2023-07-26
BTM-716: Introduce our own perf testing queue

phinterseer:


Si3rr4wow (approved):
LGTM 👍

phinterseer:
@gdsgj Yes, that's what I had in mind first. But as I explained earlier in stand-up, the perf-test queue needs to be accessible cross-account from the perf-testing account. Since queues are encrypted, I didn't want to give cross-account access to our all-purpose KMS-key but have a separate one for the perf testing queue which has a special access policy.

phinterseer:
Then all testing resources which want to use the queue would need access to that key as well. I just think it's cleaner and better divided by purpose and only in those accounts which need it.

I actually forgot the queue policy for access by the perf-test-account. Will add that now.

PR #402, opened 2023-07-28
BTM 722  overview table tests

NKannan14:
Added below additional UI tests 

- HomepageOverview table Tests:
This tests verifies that the data displayed in the Overview table on the homepage matches with data retrieved from the latest invoices per vendor from json tests data file. 
Also, added tests to verify the correct navigation to respective contract and invoice page when contract name and vie w invoice links are clicked.

- Breadcrumbs tests
Added new suite of tests to validate the breadcrumb count to ensure that the navigation breadcrumbs are working as expected.

- moved getTable function to page file so that can be reused in homepage and invoice page tests
- Added a getLatestInvoicesByVendor to get the latest invoices per vendor to compare with UI
- Added a file getPrerttyMonth the function takes month number as input and returns the corresponding short form of the month name to compare it with UI data

### DIFF
--- a/ui-tests/specs/home.spec.ts
+++ b/ui-tests/specs/home.spec.ts
@@ -26,6 +26,19 @@ describe("Home Page Tests", () => {
     );
   });
 
+  it("Should navigate to the Contracts page when clicked on the link", async () => {
+    await HomePage.clickOnContractsPageLink();
+    const newPageUrl = await browser.getUrl();
+    expect(newPageUrl).toMatch(/contracts$/);
+  });
+});
+
+describe("Home Page Overview table Tests", () => {
+  beforeEach(async () => {
+    await browser.url(" ");
+    await waitForPageLoad();
+  });
+
   it("Should display the overview table with correct data", async () => {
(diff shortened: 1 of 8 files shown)


ui-tests/specs/home.spec.ts
```diff
@@ -29,3 +32,75 @@ describe("Home Page Tests", () => {
...
+          contractName: invoice.contract_name,
+          vendor: invoice.vendor_name,
+          month: `${monthString} ${invoice.year}`,
+          reconciliationDetails:
+            reconciliationDetails.bannerMessage.toUpperCase(),
+          details: "View Invoice",
+        };
+      }
+    );
+    const sortTableDataByVendor = tableData
```
markschnitziusgovuk, line 61: (minor) Should be sortedTableDataByVendor I think.

markschnitziusgovuk (approved):
Looks fine other than that minor thing.

ui-tests/specs/home.spec.ts
NKannan14: Thanks Mark, fixed it

PR #409, opened 2023-08-01
btm-714 events missing

Xfarya:
UPDATE: testData.json had incorrect data. This file was updated - all other changes reversed.

Removed the "getPrice()" function form "billing-with-tax" string, as not needed.
This does mean that there is no string being called here - but we said this would not happen in production code (as if there is a total amount, there will be a vat amount also)


markschnitziusgovuk:
I approved, but maybe we should include the data fixes as part of this PR?

src/frontend/extract-helpers/get-reconciliation-rows.ts
```diff
@@ -44,10 +44,7 @@ export const getReconciliationRows = (
         item.billing_price_formatted,
         item.price_difference_percentage
       ),
-      billingPriceInclVat: getPrice(
-        item.billing_amount_with_tax,
-        item.price_difference_percentage
-      ),
+      billingPriceInclVat: item.billing_amount_with_tax,
```
NKannan14, line 47: When invoice data is missing, now in Invoice table Total+Vat column  is empty instead of "Invoice data missing". I think we need to be consistent in showing the message in Invoice table similar to other cell values in that table. 
image

FaryaKandC (approved):

FaryaKandC (approved):

PR #410, opened 2023-08-02
BTM 741 multi browser support

NKannan14:
This PR includes changes aimed at improve the ui test execution across different browsers with specific adjustments for safari.
1.Added wdio-safaridriver.log to .gitignore, this log file generated when test executed in safari.
2. Due to extra spaces retrieved from ui when running in safari, added .trim() to sortedTableData objects.
3. Added a function called expectEqualTrimmed to trim spaces in both actual and expected values. This was necessary because safari retrieves extra spaces, leading to assertion failures.
4. In wdio.conf added browserName as environment variable. Chrome is set as the default browser if environment variable is not set.
5. Added specific configurations for different browsers including headless options. Safari does not support headless mode, so empty options were provided for it.
6. As safari does not support parallel execution, the maxInstances for safari is limited to 1 and for other browsers the limit is 10
7. Updated readme with detailed instructions on how to run tests in different browsers.


Si3rr4wow (approved):
⭐

ui-tests/pageobjects/page.ts
```diff
@@ -11,6 +11,10 @@ export default class Page {
     return $(".govuk-heading-l");
   }
 
+  public get pageParagraphText(): Promise<WebdriverIO.Element> {
+    return $("p.govuk-body");
+  }
```
Si3rr4wow, line 16: be wary, there's likely to be ,multiple <p>s on a page

ui-tests/pageobjects/page.ts
```diff
@@ -58,6 +66,7 @@ export default class Page {
 
       tableData.push(rowData);
     }
+    console.log(tableData);
```
Si3rr4wow, line 69: best not to do this if possible, don't want to accidentally log something we shouldn't

ui-tests/pageobjects/page.ts
NKannan14: using this function on homepage and currently we have only <p> on that page

PR #411, opened 2023-08-02
BTM-713 currency fix encoding issue and optimise ui test data upload

NKannan14:
This PR addresses an issue where the currency symbol £ was being incorrectly encoded when uploading a test data file to s3. The root cause was found to be `ASCII` encoding used in putObject function.

- Modified putS3objectBasic function to accept an optional  encoding parameter. By default the encoding set to ASCII  but now it can be overridden. If UTF-8 is specified it will be used to correctly handle characters like £.
- Updated UI test data file uploading mechanism,Previously the upload was happening before each test now I changed this to onPrepare function so it is executed once.
-  Renamed `uploadExtractDataFileForUITest` function to cleanAndUploadExtractFileForUITest  and added deletion of existing file in s3 and uploading new ones.


Si3rr4wow (approved):
Good find! ⭐

PR #423, opened 2023-08-10
Reduce reserved concurrency

Si3rr4wow:
Reduces the reserved concurrency of a few resources to give us more headroom


cloudformation/frontend.yaml
```diff
@@ -137,4 +137,4 @@ Resources:
                 - athena:GetQueryExecution
                 - athena:GetQueryResults
               Resource: !Sub arn:aws:athena:${AWS::Region}:${AWS::AccountId}:workgroup/primary
-      ReservedConcurrentExecutions: 10
+      ReservedConcurrentExecutions: 2
```
Si3rr4wow, line 140: we don't have _that_ many users so I think this should be fine

PR #427, opened 2023-08-18
Longer status poll interval, switch dashboard extract func to V3

Si3rr4wow:


src/shared/utils/athenaV3.ts
```diff
@@ -7,17 +7,21 @@ import {
...
   intervalMillis: number;
+  maxAttempts: number;
 
-  MAX_ATTEMPTS = 10;
-
-  constructor(athena: AthenaClient, intervalMillis = INTERVAL_MS) {
+  constructor(
+    athena: AthenaClient,
+    intervalMillis = INTERVAL_MS,
+    maxAttempts = MAX_ATTEMPTS
```
markschnitziusgovuk, line 20: You've made maxAttempts configurable, but the only caller to this constructor doesn't set it...  Just confirming that you didn't mean to change that as well.

Si3rr4wow: Yeah it just seemed like the kinda thing there should be an option for

markschnitziusgovuk (approved):

phinterseer (approved):

PR #428, opened 2023-08-18
BTM-762 use invoice start date

markschnitziusgovuk:
We have been improperly ingesting CSV invoices using the receipt date instead of the "Invoice period start", so previously ingested invoices are (in most cases I think) saved to the wrong month on S3.

This PR fixes the code at least, so that CSVs save their data to the correct month.  There will be a later effort to fix any misfiled data.


Si3rr4wow:
Just checking there's no need to change any of the queries for this? I guess if it's all based on the folder it's in and we may just have things in the wrong folders then it's not an issue

src/shared/utils/get-standardised-invoice-key.test.ts
```diff
@@ -5,7 +5,8 @@ test("Invoice key getter with standardised line item", () => {
     invoice_receipt_id: "370 000",
     vendor_id: "vendor_testvendor1",
     vendor_name: "Vendor One",
-    invoice_receipt_date: "2023-02-28",
+    invoice_receipt_date: "2023-03-01",
```
FaryaKandC, line 8: is there a reason to keep the invoice_receipt_date (in all cases), as it is no longer being called?

Xfarya, line 8: is there a reason to keep the "invoice_receipt_date" (in all instances), as we are no longer using it?

markschnitziusgovuk:
> Just checking there's no need to change any of the queries for this? I guess if it's all based on the folder it's in and we may just have things in the wrong folders then it's not an issue


Correct, it only affects where the invoice data gets stored, and after that point, Athena uses the year and month from the folder name for each line item that gets returned from a query.

src/shared/utils/get-standardised-invoice-key.test.ts
markschnitziusgovuk: Didn't want to risk breaking things on a Friday afternoon, and it can't hurt to carry it along, so I left it in.  Even if we aren't now using it, we might want to show this value in the future.

Xfarya (approved):

PR #434, opened 2023-08-23
Enable UI Tests to run in Docker container

NKannan14:
Updated Dockerfile to install chromium and chrome driver to enable running tests in a docker container
Modified `run-tests.sh` to include the UI tests
Introduced isDocker variable to determine if the test are running in a Docker environment and conditionally load webdriver service based on the environment. Only `chromeDriver` will be loaded when running tests in Docker
Added new script in `package.json` for running tests in Docker with the IS_DOCKER=true environment variable


integration_tests/run-tests.sh
```diff
@@ -19,6 +19,12 @@ TESTS_EXIT_CODE=$?
 
 cp reports/testReport.xml $TEST_REPORT_ABSOLUTE_DIR/junit.xml
 
-if [ $TESTS_EXIT_CODE -ne 0 ]; then
+npm run test:ui
+
+TESTS_EXIT_CODE=$?
```
phinterseer, line 24: You're using the same variable twice. You need two different ones of course.

integration_tests/run-tests.sh
```diff
@@ -19,6 +19,12 @@ TESTS_EXIT_CODE=$?
 
 cp reports/testReport.xml $TEST_REPORT_ABSOLUTE_DIR/junit.xml
 
-if [ $TESTS_EXIT_CODE -ne 0 ]; then
+npm run test:ui
```
phinterseer, line 22: You might want to abort before this if the int-tests fail. But that's ultimately for you to decide if you think it's worthwhile running the ui-tests with failing int-tests.

integration_tests/run-tests.sh
NKannan14: yes thanks fixed it

integration_tests/run-tests.sh
NKannan14: Its worth to run the ui-tests even with failing int-tests as required UI data is directly uploaded to btm_extract_data folder

phinterseer (approved):
Approved assuming local testing was successful.

Si3rr4wow (approved):

PR #436, opened 2023-08-25
BTM-745: quarterly invoices part 3

gdsgj:
Shows quarterly invoices as quarterly on the website

Notes:
- renames `testData.json` to `testData.txt` since it is not a valid JSON file and so broke commits due to formatting failure
- adds caching for retrieving that file

### DIFF
--- a/src/frontend/extract-helpers/get-contract-periods.test.ts
+++ b/src/frontend/extract-helpers/get-contract-periods.test.ts
@@ -23,8 +23,6 @@ describe("getContractPeriods", () => {
         "\n" +
         fakeDataRow("2023", "03", "2", "test1") +
         "\n" +
-        fakeDataRow("2023", "04", contractId, "test1") +
-        "\n" +
         fakeDataRow("2023", "05", contractId, "test1") +
         "\n" +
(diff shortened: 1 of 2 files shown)


src/frontend/extract-helpers/get-line-items.ts
```diff
@@ -4,11 +4,18 @@ import { FullExtractLineItem } from "./types";
 export const getLineItems = async (
   contractId: string,
   year: string,
-  month: string
+  month: string | number,
+  invoiceIsQuarterly: boolean = false
 ): Promise<FullExtractLineItem[]> => {
   const dashboardData = await getDashboardExtract();
+  const monthNumber = typeof month === "number" ? month : parseInt(month, 10);
```
Si3rr4wow, line 11: When is month a number?

Si3rr4wow: I mean always obviously but I thought they were strings in the view so they could be like "08" for August for example

Si3rr4wow (approved):
Wild that you managed to do this without changing _any_ of the nujucks ⭐

PR #461, opened 2023-09-07
Fix deployment

gdsgj:
Extract function invocations queued up and so ran after UI tests started, overwriting test data

Function takes 1min in dev, so if 1min is max invocation event age, waiting 2min should work


NKannan14 (approved):
LGTM

PR #463, opened 2023-09-08
Feature/btm 773/shortfall synthetic events

markschnitziusgovuk:
Implements 'shortfall' synthetic events, which are events used to adjust counts up to a certain target, after a given month or quarter has ended.

While implementing this I refactored how the regular synthetic events work as well.  Now, whenever the lambda is called, the synthetic events config is read and processed in its entirety, looking for any month or quarter that needs synthetic events generated.  This way, we can add config that dates into the past, and the necessary events to rectify old invoices will be generated the next time the lambda runs.  This now happens for both 'shortfall' and regular ('fixed') synthetic events.

The business-logic.test.ts is large enough that github felt the need to collapse it, so please don't just skip over it.  It necessarily tests every combination of these situations:

- Event type: fixed | shortfall
- Event frequency: monthly | quarterly
- Scheduled for: past | present | future
- Already has events: none | some | all

There's maybe still some overlap with the get-active-periods.test.ts which was added later.


src/handlers/synthetic-events/business-logic.ts
```diff
@@ -1,10 +1,33 @@
 import { BusinessLogic } from "../../handler-context";
 import { Env } from "./types";
 import { ConfigElements } from "../../shared/constants";
+import { SyntheticEventDefinition } from "../../shared/types";
 import { CleanedEventBody } from "../clean/types";
 import crypto from "crypto";
 import { formatDate } from "../../shared/utils";
+import { getDashboardExtract } from "../../frontend/extract-helpers/get-dashboard-extract";
+import { FullExtractLineItem } from "../../frontend/extract-helpers/types";
```
Si3rr4wow, line 9: Bit picky but if these are used here should they be moved to the shared folder?

src/handlers/synthetic-events/business-logic.ts
```diff
@@ -13,29 +38,82 @@ export const businessLogic: BusinessLogic<
...
+  return dashboardData
+    .filter(
+      (lineItem) =>
+        lineItem.vendor_id === syntheticEventDefinition.vendor_id &&
+        eventTypes.includes(lineItem.event_name) &&
+        +lineItem.year === period.year &&
+        +lineItem.month === period.month
+    )
+    .map((lineItem) => +lineItem.transaction_quantity)
+    .reduce<number>((sum, quantity) => sum + quantity, 0);
```
Si3rr4wow, line 98:
```suggestion
  return dashboardData.reduce<number>((sum, lineItem) => {
    const shouldBeSummed =
      lineItem.vendor_id === syntheticEventDefinition.vendor_id &&
      eventTypes.includes(lineItem.event_name) &&
      +lineItem.year === period.year &&
      +lineItem.month === period.month;

    if (!shouldBeSummed) return sum;
    return sum + +lineItem.transaction_quantity;
  }, 0);
```

src/handlers/synthetic-events/business-logic.ts
```diff
@@ -13,29 +38,82 @@ export const businessLogic: BusinessLogic<
...
-        event_id: crypto.randomUUID(),
-        event_name: configLine.event_name,
-        timestamp: nowTime,
-        timestamp_formatted: nowFormatted,
-        credits: configLine.quantity,
-        component_id: configLine.component_id,
-      };
-      events.push(event);
-    }
+  syntheticEventsConfig.forEach((syntheticEventDefinition) => {
```
Si3rr4wow, line 46: This could also be a chubby reduce, doesn't save you any loops but you don't have to declare `events` and push into it 

```
  return syntheticEventsConfig.reduce<CleanedEventBody[]>(
    (acci, syntheticEventDefinition) => {
      const activePeriods = getActivePeriods(
        now,
        syntheticEventDefinition.frequency,
        syntheticEventDefinition.type,
        new Date(syntheticEventDefinition.start_date),
        syntheticEventDefinition.end_date
          ? new Date(syntheticEventDefinition.end_date)
          : undefined
      );

      return [
        ...acci,
        ...activePeriods.reduce<CleanedEventBody[]>((accj, period) => {
          const existingEventCount = countExistingEvents(
            syntheticEventDefinition,
            period,
            dashboardData
          );

          if (syntheticEventDefinition.quantity < existingEventCount)
            return accj;

          return [
            ...accj,
            createEvent(
              new Date(period.year, period.month - 1),
              syntheticEventDefinition,
              syntheticEventDefinition.quantity - existingEventCount
            ),
          ];
        }, []),
      ];
    },
    []
  );
```

PR #471, opened 2023-09-18
Fix integration test

gdsgj:
Fixes intermittent issue where checking results of test fails due to interfering invoice data from other test (in `dashboard-data-extraction-tests.ts`) with same year, month, vendor, and year


NKannan14 (approved):
LGTM

PR #481, opened 2023-09-21
Btm 832 fix teardown

phinterseer:
After going down multiple rabbit holes with this, the simplest solution is just to ignore the Athena Views on deletion of a stack as they will just be deleted as part of the database anyway...


markschnitziusgovuk (commented):
LGTM inasmuch as I understand it, but let's start having the new guys approve.

src/handlers/custom-athena-view-resource/handler.ts
phinterseer, line 1: This isn't strictly necessary to fix the issue, but I did this as part of exploring multiple options on the way and it seemed like a good idea to have the V3-Athena client and a proper physical ID for the resource.

Si3rr4wow (approved):
Thanks for doing the SDK update too 🙌

PR #487, opened 2023-09-25
BTM-827 Issue with ingesting invoices with multiple lines #486

mark-cab:


MLMartinDG (approved):

src/handlers/store-standardised-invoices/handler.ts
```diff
@@ -24,23 +26,52 @@ export const handler = async (event: SQSEvent): Promise<Response> => {
...
     throw new Error("Raw invoice bucket not set.");
 
+  for (const records of event.Records) {
+    if (!records || !records.body) {
+      throw new Error("Record contains no body");
+    }
+  }
+
+  let recordBodies: StandardisedLineItem[] = [];
+  for (const records of event.Records) {
```
MLMartinDG, line 36: Can we merge the above for loop into this one?

src/handlers/store-standardised-invoices/handler.ts
```diff
@@ -24,23 +26,52 @@ export const handler = async (event: SQSEvent): Promise<Response> => {
...
+    const recordList: StandardisedLineItem[] = JSON.parse(records.body);
+    recordBodies = recordBodies.concat(recordList);
+  }
```
Si3rr4wow, line 39:
```suggestion
  const recordBodies = flatMap(record => JSON.parse(records.body))
```

might be a more javascripty way to do this

src/handlers/store-standardised-invoices/handler.ts
mark-cab: would rather keep the validation seperate to be honest, will update second one though as it's not needed as much as it's not got lots of logs in it now

src/handlers/store-standardised-invoices/handler.ts
```diff
@@ -24,23 +26,52 @@ export const handler = async (event: SQSEvent): Promise<Response> => {
...
+      logger.error(`Handler failure for ${JSON.stringify(body)}`, { error });
+      response.batchItemFailures.push({
+        itemIdentifier: JSON.stringify(body),
+      });
     }
   });
 
   await Promise.all(promises);
+
+  if (response.batchItemFailures.length === 0) {
```
Si3rr4wow, line 60: AWS::Serverless::Api

mark-cab: not sure I follow, if some have succeeded and some have failed we don't want to move the file to the success folder we would want to reprocess it

src/handlers/store-standardised-invoices/store-line-item.test.ts
```diff
@@ -249,15 +238,15 @@ describe("Line item storer", () => {
   });
 
   test("Line item storer with record body object with no invoice-is-quarterly option", async () => {
-    delete givenRecordBodyObject.invoice_is_quarterly;
-    givenRecord = { body: JSON.stringify(givenRecordBodyObject) } as any;
+    // @ts-expect-error
+    delete givenRecord.invoice_is_quarterly;
+    givenRecord = { body: JSON.stringify(givenRecord) } as any;
```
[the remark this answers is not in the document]
mark-cab: why the ts ignore? That field is not optional but delete requires optional so when I corrected the record object to use StandardisedLineItem type this line gave a ts error

src/handlers/store-standardised-invoices/handler.ts
mark-cab: This  call is for moving the invoice file from the bucket to /successful. The files we create per line for further processing are saved as part of store-line-item.ts (putTextS3)

MLMartinDG (approved):

PR #499, opened 2023-09-29
BTM-795 event bucketing

mark-cab:


src/handlers/event-bucketing/s3connect.ts
```diff
@@ -0,0 +1,54 @@
+import {
```
MLMartinDG, line 1: There is a shared/utils/s3/ package - would it be appropriate to make some of this file a shared util?

cloudformation/event-bucketing.yaml
```diff
@@ -0,0 +1,108 @@
+AWSTemplateFormatVersion: '2010-09-09'
+Transform: AWS::Serverless-2016-10-31
+
+Resources:
+  EventBucketingFunction:
+    # checkov:skip=CKV_AWS_116: DLQ not needed for scheduled lambda that does not process events
+    Type: AWS::Serverless::Function
```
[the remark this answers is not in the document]
pskushwaha1: Lambda automatically integrates with CloudWatch Logs so it creates a default Log group  /aws/lambda/<function name> , we can create  custom Log group  for Lambda if needed , i think that is not required here ?

src/handlers/event-bucketing/s3connect.ts
mark-cab: I don't think this would add value to the utils. It's basically an interface between the utils and this lambda with some added logic specific for this lambda. Was agreed with Mark during paired programming that it's better practice to try and split this interface logic out from the handler itself.

MLMartinDG (approved):

PR #518, opened 2023-11-15
Bump chrome webdriver to 118

MLMartinDG:
Move chromedriver to 118 to get ui tests working again

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -59,7 +59,7 @@
         "aws-lambda": "^1.0.7",
         "aws-sdk-client-mock": "^2.0.0",
         "babel-jest": "^29.0.3",
-        "chromedriver": "^116.0.0",
+        "chromedriver": "^118.0.0",
         "commander": "^11.0.0",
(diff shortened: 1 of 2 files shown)


package-lock.json
```diff
@@ -8549,7 +8549,7 @@
         "chromedriver": "bin/chromedriver"
```
pskushwaha1, line 8549: Looks like the chromedriver update to 118 has some dependencies on Node Version , Not sure if this will be issue but the Test are successful

package-lock.json
```diff
@@ -8549,7 +8549,7 @@
...
       },
       "engines": {
-        "node": ">=16"
+        "node": ">=18"
```
pskushwaha1, line 8552: Looks like the chromedriver update to 118 has some dependencies on Node Version , Not sure if this will be issue but the Test are successful

MLMartinDG: It does fire a warning when it runs - the node 18 upgrade is on the roadmap.

pskushwaha1 (approved):

PR #524, opened 2023-12-14
BTM-685 Filter non DVLA if Driving license is set

MLMartinDG:
ID Events relating to Driving Licenses will be supported. For now, we want to only consider these events if the License was issued by DVLA.


mark-cab (approved):
Not a massive fan on adding event specific logic but unless we can change the data upstream guess there isn't much choice.

